PR #2, opened 2021-10-07
Initial implementation for HDFS operator

### LINKED ISSUE #1: Add HDFS service discovery code to be used by products depending on HDFS
Producst like HBase which require an underlying HDFS to work will need to look up fs strings to be used for connecting to instances of HDFS.

The code to generate these strings should be exported by this crate.

The ZooKeeper operator already has an implementation of a similar relationship and can be used as inspiration.

maltesander:
This is a first implementation for HBase.

Working:

- [x] Deployment of one name node and multiple data nodes
- [x] Monitoring for name node and data nodes
- [x] Product-config integration
- [x] Restart, Start and Stop Commands
- [x] Versioning working
- [x] Sticky scheduling
- [x] Discovery for other operators


examples/simple-hdfs-cluster.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: simple
+spec:
+  version: 3.2.2
+  nameNodes:
+    roleGroups:
+      default:
+        selector:
+          matchLabels:
+            kubernetes.io/arch: stackable-linux
```
soenkeliebau: As per https://github.com/stackabletech/issues/issues/115 we shouldn't have this selector in our examples any more and match on `kubernetes.io/os` instead.

soenkeliebau (approved):

PR #175, opened 2022-05-12
Allow external access to the HDFS cluster

### LINKED ISSUE #174: Allow external access to the HDFS cluster
Currently we just expose overlay network addresses in both hdfs-site.xml and dynamic registrations to the namenodes. This prevents users from accessing HDFS from outside of the k8s overlay network.

This requires:
 
- Namenodes must be accessed using stable addresses (such as a `LoadBalancer` `Service` where available)
  - The generated discovery profile must be altered to use this address
- Datanodes must be accessed using their associated `NodePort` `Service` addresses

Datanode addresses cannot currently be overridden in HDFS, I have outlined the changes required for this in https://github.com/stackabletech/hadoop/tree/spike/override-datanode-id. This is a purely server-side change.

I have a spike that implements this semi-manually in https://github.com/stackabletech/hdfs-operator/tree/spike/external-access, if we consider that access pattern acceptable then we can start looking into how to integrate it better into both HDFS and the operator.

nightkr:
## Description

Fixes #174

This allows users to access the HDFS cluster from outside of the K8s cluster. This requires a number of changes:

- We configure the discovery configmap to use `LoadBalancer` services for access rather than the namenode pod hostnames
  - Prefer LB over NodePort services since k8s nodes are often temporary and fungible in cloud environments (especially when using managed K8s services)
  - We use a LB per namenode to allow clients to use their regular failover logic when namenodes are down/standby
- Configure the datanodes to register their NodePort connection info rather than their Pod IP
  - This depends on https://issues.apache.org/jira/browse/HDFS-16577
  - NodePort is fine here since clients fetch this connection info dynamically from the namenodes

## Review Checklist


examples/simple-hdfs-cluster.yaml
```diff
@@ -65,3 +65,62 @@ spec:
...
+  selector:
+    statefulset.kubernetes.io/pod-name: simple-hdfs-namenode-default-1
+  ports:
+    - name: http
+      port: 9870
+      targetPort: http
+    - name: rpc
+      port: 8020
+      targetPort: rpc
+---
```
nightkr: Need to set up this RBAC to be more fine-grained, like we do in the Kafka operator

rust/crd/src/lib.rs
```diff
@@ -384,19 +385,23 @@ impl HdfsCluster {
...
+    pub external_name: Option<String>,
 }
 
 impl HdfsPodRef {
     pub fn fqdn(&self) -> String {
-        format!(
-            "{}.{}.{}.svc.cluster.local",
-            self.pod_name, self.role_group_service_name, self.namespace
-        )
+        self.external_name.clone().unwrap_or_else(|| {
```
nightkr, line 523: Not super happy with abusing the FQDN property here (or `HdfsPodRef` at all..?), this also breaks down if the user overrides the ports for whatever reason.

rust/operator/src/hdfs_controller.rs
```diff
@@ -472,10 +559,16 @@ fn datanode_containers(
     vec![Container {
         name: rolegroup_ref.role.clone(),
         args: Some(vec![
-            format!("{hadoop_home}/bin/hdfs", hadoop_home = HADOOP_HOME),
-            "--debug".to_string(),
-            "datanode".to_string(),
+            "sh".to_string(),
+            "-c".to_string(),
+            format!(
+                r#"kubectl get svc $POD_NAME -o json > /data/pod-svc && kubectl get node $NODE_NAME -o json > /data/pod-node && DATA_PORT=$(jq '.spec.ports[] | select(.name == "data") | .nodePort' /data/pod-svc) HTTP_PORT=$(jq '.spec.ports[] | select(.name == "http") | .nodePort' /data/pod-svc) IPC_PORT=$(jq '.spec.ports[] | select(.name == "ipc") | .nodePort' /data/pod-svc) NODE_IP=$(jq -r '.status.addresses | map(select(.type == "ExternalIP")) | .[0].address' /data/pod-node) {HADOOP_HOME}/bin/hdfs --debug datanode"#
```
nightkr: Need to clean this up better

rust/crd/src/lib.rs
nightkr: @razvan would it make sense to split `HdfsPodRef` into separate types for journalnodes and namenodes? We'll probably want to only use the external name for *all* namenode traffic anyway, once we start doing Kerberos (since this is linked to the hostname, and HDFS nodes dislike having more than one Kerberos principal each).

razvan: Makes sense.

nightkr:
This has ended up becoming an example use case of https://github.com/stackabletech/lb-operator/pull/1, and merging this is blocked on releasing that.

maltesander (commented):
Need to add some rolebindings:
```
rolebindings.rbac.authorization.k8s.io "hdfs-datanode-55df4c3b-06cc-4311-b2a7-86bfd5d2162f" is forbidden: User "system:serviceaccount:default:hdfs-operator-serviceaccount" cannot patch resource "rolebindings" in API group "rbac.authorization.k8s.io" in the namespace "default": Forbidden
```

PR #290, opened 2022-12-30
[Merged by Bors] - Enable Log Aggregation for HDFS

maltesander:
# Description

- Enable logging for namenode, datanode, journalnode containers as well as the namenode side container ZKFC.
- Snafufied the whole operator
- Refactored container handling
- Local integration tests work

fixes https://github.com/stackabletech/hdfs-operator/issues/288

test: https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/49/

## Review Checklist

- [x] Code contains useful comments
- [x] CRD change approved (or not applicable)
- [x] (Integration-)Test cases added (or not applicable)
- [x] Documentation added (or not applicable)
- [x] Changelog updated (or not applicable)
- [x] Cargo.toml only contains references to git tags (not specific commits or branches)
- [x] Helm chart can be installed and deployed operator works (or not applicable)

Once the review is done, comment `bors r+` (or `bors merge`) to merge. [Further information](https://bors.tech/documentation/getting-started/#reviewing-pull-requests)

PR #291, opened 2023-01-02
Multiple datanode pvcs

### LINKED ISSUE #274: Support for multiple storage directories
Currently, only a single storage directory is supported. We need to support an arbitrary number.
While we're also want to support heterogenous storage tagging.
CRD design
```
  dataNodes:
    roleGroups:
      default:
        replicas: 1
        config:
          resources:
            storage:
              data: # We need to overwrite the data pvcs coming from the default value
                count: 0
              my-disks:
                capacity: 100Gi
                count: 5
                hdfsStorageType: Disk
              my-ssds:
                capacity: 10Gi
                storageClass: premium
                count: 3
                hdfsStorageType: Ssd
```

By defaulting to 
```
              data
                capacity: 5Gi
                count: 1
                hdfsStorageType: Disk
```
this is a non-breaking change for the CRD. Because we name the pvcs `data`, `data-1`, `data-2` etc. even the pvc names do not change, so the whole feature should be non-breaking

- [ ] Multiple pvcs supported
- [ ] `hdfsStorageType` is passed to hdfs
- [ ] CRD change is non-breaking
- [ ] pvc name change is non-breaking
- [ ] Tests added
- [ ] Documentation added

sbernauer:
# Description

Fixes https://github.com/stackabletech/hdfs-operator/issues/274

## Review Checklist

- [ ] CRD change approved (or not applicable)


sbernauer:
Started test https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/37/

sbernauer:
Tests passed :sweat_smile:

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -169,6 +169,47 @@ In the above example, all data nodes in the default group will store data (the l
...
+              count: 2
+              hdfsStorageType: Disk
+            ssd:
+              capacity: 500Gi
+              storageClass: premium
+              count: 1
+              hdfsStorageType: Ssd
+----
+
+This will create 3 pvcs:
```
maltesander:
```suggestion
This will create 3 PVCs:
```

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -169,6 +169,47 @@ In the above example, all data nodes in the default group will store data (the l
 
 By default, in case nothing is configured in the custom resource for a certain role group, each Pod will have a `1Gi` large local volume mount for the data location.
 
+==== Multiple storage volumes
+
+Datanodes can have multiple disks attached to increase the storage size as well as speed.
+They can be of different type, e.g. HDDs or SSDs.
+
+You can configure multiple PersistentVolumeClaims (pvcs) for the datanodes as follows:
```
maltesander: I would capitalize "pvc".
```suggestion
Datanodes can have multiple disks attached to increase the storage size as well as speed.
They can be of different type, e.g. HDDs or SSDs.

You can configure multiple [PersistentVolumeClaims](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) (PVCs) for the datanodes as follows:
```

rust/crd/src/storage.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "PascalCase")]
+pub enum HdfsStorageType {
+    Archive,
+    Disk,
+    Ssd,
+    RamDisk,
+}
```
maltesander: Just a personal opinion. I would still go for CamelCase or upper case here (that would match the way HDFS expects it). `Ssd` looks very strange.

rust/crd/src/storage.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// There are two options when it comes to naming:
+    /// 1. Name the pvcs `data-0, data-1, data-2, ... , data-{n-1}`
+    /// ** Good, because consistent naming
+    /// ** Bad, because existing deployments (using release 22.11 or earlier) will need to migrate their data by renaming their pvcs
+    ///
+    /// 2. Name the pvcs `data, data-1, data-2, ... , data-{n-1}`
+    /// ** Good, if nodes only have a single pvc (which probably most of the deployments will have) they name of the pvc will be consistent with the name of all the other pvcs out there
+    /// ** It is important that the first pvc will be called `data` (without suffix), regardless of the number of pvcs to be used. This is needed as nodes should be able to alter the number of pvcs attached and the use-case number of pvcs 1 -> 2 should be supported without renaming pvcs.
+    ///
+    /// This function uses the 2. option.
```
maltesander, line 160: We went for option 1 here? I would just add why and not list both options.

maltesander (commented):
LGTM overall. https://github.com/stackabletech/hdfs-operator/pull/290 will change a lot of things regarding containers, errors etc. 
Since these are fewer changes i would wait for #290 to be merged first.

rust/crd/src/storage.rs
sbernauer: We actually went for Option 2 because Option 1 has
`** Bad, because existing deployments (using release 22.11 or earlier) will need to migrate their data by renaming their pvcs`
I personally like the "ADR-style" of also including discarded options and why there were discarded

rust/crd/src/storage.rs
sbernauer: Sorry, didn't get this. Are you suggesting to rename `Ssd` -> `SSD`? Just wanted to mention [this k8s docs](https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects)
I can remember a discussion with Teo where he wanted to stick to strict PascalCase but i can be wrong here

maltesander: Ok i do not remember or was not part of that discussion.
What i mean is either go for a `rename_all = "CamelCase"` which would lead to `archive`, `disk`, `ssd`, `ramDisk` or go all upper for `SSD` and so on (only serialization / deserilization, not the enum field). I feel `Ssd` just looks so wrong.
If you feel this was already decided just ignore this.

sbernauer: Update the PR to use `Archive/Disk/SSD/RAMDisk`. Let's revisit when the poll results are there

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -169,6 +169,47 @@ In the above example, all data nodes in the default group will store data (the l
...
+
+1. `data-hdfs-datanode-default-0` (5Ti)
+2. `data-1-hdfs-datanode-default-0` (5Ti)
+3. `ssd-hdfs-datanode-default-0` (500Gi)
+
+By configuring and using a dedicated `storageClass` you can configure your HDFS to use local disks attached to Kubernetes nodes.
+
+[NOTE]
+====
+You might need to re-create the StatefulSet to apply the new pvc configurations because of https://github.com/kubernetes/kubernetes/issues/68737[this Kubernetes issue].
```
maltesander:
```suggestion
You might need to re-create the StatefulSet to apply the new PVC configuration because of https://github.com/kubernetes/kubernetes/issues/68737[this Kubernetes issue].
```

rust/crd/src/storage.rs
sbernauer: 0 for `PascalCase everything: Archive/Disk/Ssd/RamDisk`
5 for `Try to stick to k8s convention, which seems to uppercase abbreviations Archive/Disk/SSD/RAMDisk`
2 fir `Try to stick to HDFS literals ARCHIVE/DISK/SSD/RAM_DISK`

sbernauer:
Superseded by https://github.com/stackabletech/hdfs-operator/pull/296

Issue #264, opened 2022-11-01
Namenodes deadlock in safemode when all namenodes have been down at the same time


nightkr:
Because we set `dfs.ha.nn.not-become-active-in-safemode`, each namenode starts in safemode until it has:

1. Verified some amount of datanode blocks
2. Downloaded the latest fsimage from another active namenode

If all namenodes go down at the same time then condition 2 will never be met, since there is no active namenode to synchronize with.

Unsetting `dfs.ha.nn.not-become-active-in-safemode` fixes this by allowing a safemode NN to take over, but the implications of this should probably be investigated by someone more familiar with HDFS first. Presumably we set it for a reason in the first place.

lfrancke:
This was set by accident at the very beginning of our operator due to a misinterpretation of the [docs](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/HDFSHighAvailabilityWithNFS.html) of HDFS.

PR #334, opened 2023-03-16
[Merged by Bors] - Allow enabling secure mode with Kerberos

### LINKED ISSUE #178: Allow enabling secure mode with Kerberos on HDFS
As a user/admin I'd like to be able to enable secure mode in HDFS using Kerberos so I can authenticate all requests against HDFS.

This is done when
- [x] a CRD change(s) has been proposed and reviewed on how to enable Kerberos
- [x] it uses the functionality provided by the secret operator to provision Kerberos keytabs https://github.com/stackabletech/secret-operator/issues/4
- [x] documentation is added
- [x] integration tests are added
- [ ] examples are added
- [x] feature tracker is updated

### LINKED ISSUE #338: Cluster bricks when all journalnodes are down
### Affected version

main

The JournalNodes seem to get stuck in a crashloop if all are deleted at the same time, complaining about not being able to find missing edits.

### Possible solution

_No response_

### Additional context

_No response_

### Environment

_No response_

None

sbernauer:
# Description

Closes https://github.com/stackabletech/hdfs-operator/issues/178
Fixes https://github.com/stackabletech/hdfs-operator/issues/338

TODOs

- [x] Release new Hadoop image with openssl and Kerberos clients use in docs and tests
- [x] Release and use operator-rs change
- [x] Fix hardcoded `kinit nn/simple-hdfs-namenode-default.default.svc.cluster.local@CLUSTER.LOCAL -kt /stackable/kerberos/keytab` in entrypoints
- [x] Go through all hadoop settings and see if they can be improved
- [X] Test different realms
- [x] Discuss CRD change
- [x] Discuss how to expose this in Discovery CM -> During on-site 2023/05 we have decided to ship this feature without exposing it via discovery *for now*
- [x] Implement discovery
- [x] Tests
- [x] Docs
- [x] Let  @maltesander have a look how we can better include the init container in the code structure
- [x] Test long running cluster (maybe turn down ticket lifetime for that)

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Documentation added or updated
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


Cargo.toml
```diff
@@ -4,6 +4,6 @@ members = [
     "rust/crd", "rust/operator", "rust/operator-binary"
 ]
 
-# [patch."https://github.com/stackabletech/operator-rs.git"]
+[patch."https://github.com/stackabletech/operator-rs.git"]
 # stackable-operator = { path = "/home/sbernauer/stackabletech/operator-rs" }
-# stackable-operator = { git = "https://github.com/stackabletech//operator-rs.git", branch = "feat/affinities" }
+stackable-operator = { git = "https://github.com/stackabletech//operator-rs.git", branch = "main" }
```
nightkr: Remove once op-rs is released

deploy/helm/hdfs-operator/crds/crds.yaml
```diff
@@ -34,6 +34,40 @@ spec:
...
+                      properties:
+                        kerberosSecretClass:
+                          default: kerberos
+                          description: Name of the SecretClass providing the keytab for the HDFS services.
+                          type: string
+                        requestNodePrincipals:
+                          default: false
+                          description: Wether a principal including the Kubernetes node name should be requested. The principal could e.g. be `HTTP/my-k8s-worker-0.mycorp.lan`. This feature is disabled by default, as the resulting principals can already by existent e.g. in Active Directory which can cause problems.
+                          type: boolean
+                        tlsSecretClass:
```
nightkr: I'm not sure, I could see an argument for creating a `security` supergroup since this isn't strictly about Kerberos... On the other hand, how useful is it outside of that context?

deploy/helm/hdfs-operator/crds/crds.yaml
```diff
@@ -34,6 +34,40 @@ spec:
...
+                          default: tls
+                          description: Name of the SecretClass providing the tls certificates for the WebUIs.
+                          type: string
+                        wireEncryption:
```
nightkr: If you want to shoot yourself in the foot then `configOverrides` already exists. No need for us to encourage this insanity.

docs/modules/hdfs/examples/getting_started/zk.yaml
```diff
@@ -10,4 +10,4 @@ spec:
   servers:
     roleGroups:
       default:
-        replicas: 1
+        replicas: 3
```
nightkr: I'm assuming this is not intended to be merged?

rust/operator/src/container.rs
```diff
@@ -483,42 +585,117 @@ impl ContainerConfig {
...
+            namespace = hdfs.namespace().context(ObjectHasNoNamespaceSnafu)?,
+        );
+        Ok(format!("echo \"Getting ticket for {principal}\" from /stackable/kerberos/keytab && kinit \"{principal}\" -kt /stackable/kerberos/keytab"))
+    }
+
+    // Command to export `KERBEROS_REALM` env var to default real from krb5.conf, e.g. `CLUSTER.LOCAL`
+    fn export_kerberos_real_env_var_command() -> String {
+        "export KERBEROS_REALM=$(grep -oP 'default_realm = \\K.*' /stackable/kerberos/krb5.conf)"
+            .to_string()
+    }
```
nightkr: I don't _think_ this dance should be required, if you try to `kinit` without a realm it should guess the realm automatically.

rust/operator/src/container.rs
```diff
@@ -623,6 +824,15 @@ impl ContainerConfig {
             .build(),
         ];
 
+        // Adding this for all containers, as not only the main container needs Kerberos or TLS
+        if hdfs.has_kerberos_enabled() {
+            volume_mounts.push(VolumeMountBuilder::new("kerberos", "/stackable/kerberos").build());
+        }
+        if hdfs.https_secret_class().is_some() {
```
nightkr:
```suggestion
        if hdfs.has_https_enabled() {
```

rust/operator/src/container.rs
```diff
@@ -724,8 +930,15 @@ impl ContainerConfig {
                     format!(
                         "-javaagent:/stackable/jmx/jmx_prometheus_javaagent-0.16.1.jar={metrics_port}:/stackable/jmx/{role}.yaml",
                     )];
-                if let Some(Some(memory_limit)) =
-                    resources.limits.as_ref().map(|limits| limits.get("memory"))
+
+                if hdfs.has_kerberos_enabled() {
+                    jvm_args.push(
+                        "-Djava.security.krb5.conf=/stackable/kerberos/krb5.conf".to_string(),
```
nightkr, line 929: Do we want to factor this out, or respect `HADOOP_OPTS` somehow?

rust/operator/src/container.rs
```diff
@@ -724,8 +930,15 @@ impl ContainerConfig {
...
+                    );
+                }
+
+                if let Some(Some(Some(memory_limit))) =
+                    resources.map(|r| r.limits.as_ref().map(|limits| limits.get("memory")))
```
nightkr: Nit:

```suggestion
                if let Some(memory_limit) =
                    resources.and_then(|r| r.limits.as_ref()?.get("memory"))
```

rust/operator/src/hdfs_controller.rs
```diff
@@ -415,10 +428,69 @@ fn rolegroup_config_map(
                 core_site_xml = CoreSiteConfigBuilder::new(hdfs_name.to_string())
                     .fs_default_fs()
                     .ha_zookeeper_quorum()
+                    .kerberos_config(hdfs, role, hdfs_name, &hdfs_namespace)
                     // the extend with config must come last in order to have overrides working!!!
                     .extend(config)
                     .build_as_xml();
             }
+            // PropertyNameKind::File(file_name) if file_name == HADOOP_POLICY_XML => {
```
nightkr: Legacy/testing stuffs?

rust/operator/src/kerberos.rs
```diff
@@ -0,0 +1,139 @@
...
+        hdfs_name: &str,
+        hdfs_namespace: &str,
+    ) -> &mut Self {
+        if let Some(kerberos_config) = hdfs.kerberos_config() {
+            self
+                .add("hadoop.security.authentication", "kerberos")
+                .add("hadoop.security.authorization", "true")
+                // Otherwise we fail with `java.io.IOException: No groups found for user nn`
+                // Default value is `dr.who=`, so we include that here
+                .add("hadoop.user.group.static.mapping.overrides", "dr.who=;nn=;nm=;jn=;")
```
nightkr: Note to self that this is a placeholder until we have group mapping (https://github.com/stackabletech/opa-operator/pull/433)

tests/templates/kuttl/kerberos/02-create-kerberos-secretclass.yaml.j2
```diff
@@ -0,0 +1,72 @@
...
+data:
+{% if test_scenario['values']['kerberos-backend'] == 'mit' %}
+  # To create keytab. When promted enter password asdf
+  # cat | ktutil << 'EOF'
+  # list
+  # add_entry -password -p stackable-secret-operator@CLUSTER.LOCAL -k 1 -e aes256-cts-hmac-sha384-192
+  # wkt /tmp/keytab
+  # EOF
+{% if test_scenario['values']['kerberos-realm'] == 'CLUSTER.LOCAL' %}
+  keytab: BQIAAABdAAEADUNMVVNURVIuTE9DQUwAGXN0YWNrYWJsZS1zZWNyZXQtb3BlcmF0b3IAAAABZAYWIgEAFAAgm8MCZ8B//XF1tH92GciD6/usWUNAmBTZnZQxLua2TkgAAAAB
```
nightkr, line 57: Probably want to decide how to handle these before merging

docs/modules/hdfs/examples/getting_started/zk.yaml
sbernauer: Not entirely sure tbh ^^ Let's remove it again

rust/operator/src/hdfs_controller.rs
sbernauer: We did run into the error mentioned in comments. I have no idea what I did to resolve them, but it seemed to have worked. Let's remove them, we can always get them from the history again

rust/operator/src/container.rs
sbernauer: We do not only use this in `kinit`, but also for
`.add("hadoop.registry.kerberos.realm", "${env.KERBEROS_REALM}")`

Cargo.toml
sbernauer: Removed

deploy/helm/hdfs-operator/crds/crds.yaml
sbernauer: Asked in Slack: https://stackable-workspace.slack.com/archives/C031A56R127/p1684218706958919

rust/operator/src/container.rs
sbernauer: I have to admit, I didn't fully get that.
Just for context, this is the reason we split HADOOP_OPTS and HDFS_NAMENODE_OPTS https://github.com/stackabletech/hdfs-operator/blob/06ca693c1205b2474fd24a3d4ab5a7a37b3d20cb/rust/operator/src/container.rs#L662-L695

rust/operator/src/container.rs
nightkr: Fair.. This still feels hacky compared to parsing the file properly, but it's not like krb5 ships a better tool for it, and "use Python (or whatever) to FFI-call `krb5_get_default_realm`" sucks too.

rust/operator/src/container.rs
nightkr: Yes, but.. this will break existing users who already rely on setting a `HADOOP_OPTS` override, in the sense that their settings will no longer take effect, right? Unless I'm missing something here.

rust/operator/src/container.rs
sbernauer: Agreed

rust/operator/src/container.rs
sbernauer: If a users currently overrides HADOOP_OPTS this override will still take precedence, so it should not break any existing setup. The implications is that some parts might not work because the miss the `java.security.krb5.conf` setting *when enabling kerberos*.
But i think this is the usual risk when using overrides.

tests/templates/kuttl/kerberos/02-create-kerberos-secretclass.yaml.j2
sbernauer: Do you see any room for improvement? I personally would be fine with merging as-is, but also happy about improvement ideas :see_no_evil:

deploy/helm/hdfs-operator/crds/crds.yaml
sbernauer: So you suggest something like the following?
```
      spec:
        clusterConfig:
          security:
            kerberos:
              tlsSecretClass: tls
              kerberosSecretClass: kerberos-$NAMESPACE
              wireEncryption: {{ test_scenario['values']['wire-encryption'] }}
```
I would say all these settings are security-related, yes. But for other product we have spec.clusterconfig.{authentication,authorization}, without putting it below spec.clusterconfig.security. So I guess the current way is more konsitent, WDYT?

rust/operator/src/container.rs
nightkr: OK.

deploy/helm/hdfs-operator/crds/crds.yaml
nightkr: More like:

```yaml
      spec:
        clusterConfig:
          security:
            tlsSecretClass: tls
            kerberos:
              kerberosSecretClass: kerberos-$NAMESPACE
            wireEncryption: {{ test_scenario['values']['wire-encryption'] }}
```

Since TLS (for example) doesn't have much to do with Kerberos compared to any other hypothetical authn method. Then again, there's the question of how likely that is to happen anyway...

rust/operator/src/container.rs
nightkr: https://github.com/stackabletech/docker-images/issues/371

nightkr: I think this is ok for now, until we solve that issue.

tests/templates/kuttl/kerberos/02-create-kerberos-secretclass.yaml.j2
nightkr: I think the ideal would be to generate the keytabs on demand, but that also brings downsides.. bleh

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
+= Security
+
+== Authentication
+Currently the only authentication mechanism support is Kerberos, which is disabled by default.
```
nightkr:
```suggestion
Currently the only supported authentication mechanism is Kerberos, which is disabled by default.
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+For Kerberos to work a Kerberos server is needed, which the users needs to provide.
```
nightkr:
```suggestion
For Kerberos to work a Kerberos KDC is needed, which the users needs to provide.
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+The xref:home:secret-operator:secretclass.adoc#backend-kerberoskeytab[secret-operator documentation] states which kind of Kerberos servers are supported and how they can be configured.
+Our integration tests use a mit krb5 server which is installed on Kubernetes, but is not production ready, e.g. because no high availability is configured.
```
nightkr:
```suggestion
Our integration tests use a MIT krb5 server which is installed on Kubernetes. Our test setup is, however, not production ready. For example, it does not configure any high availability.
```

nightkr: Honestly I'd almost strike this whole line entirely.

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== 1. Prepare Kerberos server
+To configure HDFS to use Kerberos you first need to collect information about your Kerberos server, e.g. hostname and port.
+Additionally you need a service-user, which the secret-operator uses to create create principals for the HDFS services.
+
+=== 2. Create SecretClass
```
nightkr:
```suggestion
=== 2. Create Kerberos SecretClass
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+    kerberos:
+      tlsSecretClass: tls # Optional, defaults to "tls"
+      kerberosSecretClass: kerberos-hdfs # Put your SecretClass name in here
+----
+
+The `kerberosSecretClass` is used to give HDFS the possibility to request keytabs from the secret-operator.
+
+The `tlsSecretClass` is needed to request TLS certificates, used e.g. for the Web UIs.
+
+As an alternative you can
```
nightkr:
```suggestion

```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== 4. Verify Kerberos is used
```
nightkr:
```suggestion
=== 4. Verify that Kerberos is used
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+One option is to use `stackablectl services list --all-namespaces` to get the endpoint the HDFS namenodes are reachable.
```
nightkr:
```suggestion
Use `stackablectl services list --all-namespaces` to get the endpoints where the HDFS namenodes are reachable.
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+Open the link (note that the namenode is now using https).
+You should see a Web UI similar to the following:
+
+image:hdfs_webui_kerberos.png[]
+
+The important part is
+
+> Security is on.
+
+An alternative is to shell into the namenode and try to access the file system.
```
nightkr:
```suggestion
You can alsoshell into the namenode and try to access the file system.
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+`kubectl exec -it hdfs-namenode-default-0 -c namenode -- bash -c 'bin/hdfs dfs -ls /'`
+
+You should get a `org.apache.hadoop.security.AccessControlException: Client cannot authenticate via:[TOKEN, KERBEROS]`.
```
nightkr:
```suggestion
You should get the error message `org.apache.hadoop.security.AccessControlException: Client cannot authenticate via:[TOKEN, KERBEROS]`.
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== 5. Access HDFS
+In case you want to access your HDFS it is recommended to start up a client Pod that connects to HDFS over shelling into the namenode.
```
nightkr:
```suggestion
In case you want to access your HDFS it is recommended to start up a client Pod that connects to HDFS, rather than shelling into the namenode.
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+|Authentication
+|Establishes mutual authentication between the client and the server.
+ Sets `hadoop.rpc.protection` to `authentication`, `hadoop.data.transfer.protection` to `authentication` and `dfs.encrypt.data.transfer` to `false`.
+
+|Integrity
+|In addition to Authentication, it guarantees that a man-in-the-middle cannot tamper with messages exchanged between the client and the server.
+Sets `hadoop.rpc.protection` to `integrity`, `hadoop.data.transfer.protection` to `integrity` and `dfs.encrypt.data.transfer` to `false`.
+
+|Privacy
```
nightkr:
```suggestion
|Privacy (default)
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -0,0 +1,109 @@
...
+You can specify the wire encryption mode to use as follows:
+
+[source,yaml]
+----
+spec:
+  clusterConfig:
+    kerberos:
+      # kerberosSecretClass: kerberos
+      wireEncryption: Privacy
+----
```
nightkr:
```suggestion
It is strongly recommended to leave `wireEncryption` in its default `Privacy` mode, since `Authentication` and `Integrity` leave the system vulnerable to man-in-the-middle and traffic sniffing attacks, respectively.
```

tests/test-definition.yaml
```diff
@@ -2,12 +2,12 @@
 dimensions:
   - name: hadoop
     values:
-      - 3.2.2-stackable23.4.0-rc3
-      - 3.3.3-stackable23.4.0-rc3
-      - 3.3.4-stackable23.4.0-rc3
+      # - 3.2.2-stackable0.0.0-dev
+      # - 3.3.3-stackable0.0.0-dev
```
nightkr: Are we dropping support for these?

tests/test-definition.yaml
```diff
@@ -35,31 +40,33 @@ dimensions:
...
-      - zookeeper
-      - number-of-datanodes
-      - datanode-pvcs
+  # - name: smoke
+  #   dimensions:
+  #     - hadoop
+  #     - zookeeper
+  #     - number-of-datanodes
+  #     - datanode-pvcs
+  #     - listener-class
```
nightkr: Reenable before merging??

tests/test-definition.yaml
```diff
@@ -35,31 +40,33 @@ dimensions:
...
+  #     - hadoop-latest
+  #     - zookeeper-latest
+  # - name: logging
+  #   dimensions:
+  #     - hadoop
+  #     - zookeeper-latest
+  # - name: cluster-operation
+  #   dimensions:
+  #     - hadoop-latest
+  #     - zookeeper-latest
```
nightkr: Reenable before merging?

docs/modules/hdfs/pages/usage-guide/security.adoc
sbernauer: Changed to `You can also shell into the namenode and try to access the file system:`

docs/modules/hdfs/pages/usage-guide/security.adoc
sbernauer: striked :heavy_check_mark:

tests/templates/kuttl/kerberos/02-create-kerberos-secretclass.yaml.j2
sbernauer: Yeah, it's tricky...

tests/test-definition.yaml
sbernauer: reenabled

tests/test-definition.yaml
sbernauer: reenabled

tests/test-definition.yaml
sbernauer: I only removed `3.3.3` in favor of `3.3.4` to keep the number of tests manageable

docs/modules/hdfs/pages/usage-guide/security.adoc
sbernauer: I would prefer to document the wire-encryption setting. WDYT of

In case kerberos is enabled, the default value is `Privacy`, since `Authentication` and `Integrity` leave the system vulnerable to man-in-the-middle and traffic sniffing attacks, respectively.
The security measures come with a cost of a potentially degraded performance, thus wire encryption can be disabled, but is recommend to be enabled for security reasons.

You can specify the wire encryption mode to use as follows:
[...]

deploy/helm/hdfs-operator/crds/crds.yaml
sbernauer: Implemented in 4a2afb92659f448ee25e2290459a67e73afebe6c

docs/modules/hdfs/pages/usage-guide/security.adoc
nightkr: I definitely don't think it's appropriate to have language that could encourage users to touch it in a guide document like this.

sbernauer: IMHO it's a totally legit use-case to enable Kerberos but don't use wire-encryption (I would say every hdfs clusters I have used so far had this setup (if Kerberos was used at all))

sbernauer: Sorry, you changed your comment. Do you have an idea what can we add to make it more even more clear to not disable wire encryption?

nightkr: > IMHO it's a totally legit use-case to enable Kerberos but don't use wire-encryption (I would say every hdfs clusters I have used so far had this setup (if Kerberos was used at all))

I think this is the crux of the disagreement though. What's the scenario where you do care about authentication but not about having it be useful for anything?

tests/test-definition.yaml
nightkr: OK.

nightkr (commented):
LGTM, modulo the questions that are still open.

docs/modules/hdfs/pages/usage-guide/security.adoc
sbernauer: let me cheekily copy Lars' words:

> Kerberos itself is a secure authentication protocol already so it is independent of wire encryption and there might be use-cases where people want to have best possible performance and don't care about whether anyone can read the data on the network (e.g. other protection mechanisms)

sbernauer: Removed the whole wire-encryption thing

deploy/helm/hdfs-operator/crds/crds.yaml
sbernauer: We have decided on new structure in Arch meeting

deploy/helm/hdfs-operator/crds/crds.yaml
sbernauer: Removed the whole wire-encryption thing

sbernauer:
@nightkr adopted the CRD to the Arch meeting decision. Removed the wire-encryption entirely.
Please have another look!

nightkr (approved):

rust/operator/src/kerberos.rs
sbernauer: Closing, as this was a note to self

tests/templates/kuttl/kerberos/02-create-kerberos-secretclass.yaml.j2
sbernauer: I think we did our reasonable best in the PR, I think further improvements would be pretty complicated and could be carried out in secret-op tests first

sbernauer:
Started test https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/72/
I will also spin up an HDFS and have a look at it tomorrow

sbernauer:
Next try: https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/73

sbernauer:
Tests are green again.
Long-running hdfs clusters are fine as well, but well have to wait a bit more
```
Default principal: nn/hdfs-namenode-default.kuttl-test-adjusted-oarfish.svc.cluster.local@PROD.MYCORP

Valid starting     Expires            Service principal
05/24/23 13:57:11  05/25/23 13:57:11  krbtgt/PROD.MYCORP@PROD.MYCORP
```

sbernauer:
All my namenodes are crashing with
`org.apache.hadoop.hdfs.server.namenode.EditLogInputException: Error replaying edit log at offset 0.  Expected transaction ID was 1`
`Caused by: org.apache.hadoop.hdfs.server.namenode.RedundantEditLogInputStream$PrematureEOFException: got premature end-of-file at txid 0; expected file to go up to 10 `

I guess the reason is that secret-op writes an annotation `restarter.stackable.tech/expires-at.423b4dce15bb8d736818cf7067bc2d91: "2023-05-26T12:06:22.264128554+00:00"` because we get a TLS cert from it with a lifetime of 24h.

I don't *think* this is a Kerberos problem, but we are propably running into https://github.com/stackabletech/hdfs-operator/issues/338.
This causes real problems, as all journalnodes will restart at the same time thus bricking the HDFS :(
And if this does not cause problems, https://github.com/stackabletech/hdfs-operator/issues/264 will propably.

CC @lfrancke for prioritization of https://github.com/stackabletech/hdfs-operator/issues/338

nightkr:
See also: https://github.com/stackabletech/secret-operator/issues/159. Arguably we should fix both that and #338.

sbernauer:
Just for the record: Maybe PodDisruptionBudgets could be a quick win. I think evictions (from commons-op) should respect that, but I'm only on my phone

sbernauer:
Quick update: I'm able to reproduce the problem when execution `kubectl -n kuttl-test-pumped-maggot delete pod hdfs-journalnode-default-0 hdfs-journalnode-default-1 hdfs-journalnode-default-2`
Strangely if I do this with non-kerberized Cluster everything is fine :thinking: (This clusters even survive `kubectl -n kuttl-test-fitting-silkworm delete pod hdfs-zk-server-default-0 hdfs-zk-server-default-1 hdfs-zk-server-default-2 hdfs-journalnode-default-0 hdfs-journalnode-default-1 hdfs-journalnode-default-2 hdfs-namenode-default-0 hdfs-namenode-default-1 hdfs-datanode-default-0 hdfs-datanode-default-1`)

rust/operator/src/container.rs
```diff
@@ -391,26 +482,31 @@ impl ContainerConfig {
...
+                        ACTIVE_NAMENODE=$namenode_id
                         echo "active"
                         break
                       fi
                       echo ""
                     done
 
                     set -e
-                    if [ ! -f "{NAMENODE_ROOT_DATA_DIR}/current/VERSION" ]
+                    if ! ls {NAMENODE_ROOT_DATA_DIR}/current/fsimage_*
```
sbernauer: Maybe this causes the JN problems. @nightkr *only if you have the time and inclination*: This is from your spike, what was the intention of changing this?
Will give the old code a test and see if the problem persists

sbernauer: Ok, didn't change anything :/

soenkeliebau: Would probably cause breakage when upgrading the operator later on though, similar to when we renamed datanode paths.

sbernauer: Would it? This only changes the command in the init container

sbernauer: This change is actually correct and needed. Feel free to reopen if desired!

sbernauer:
Started test https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/76/

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -43,7 +45,7 @@ The important part is
 > Security is on.
 
 You can also shell into the namenode and try to access the file system:
-`kubectl exec -it hdfs-namenode-default-0 -c namenode -- bash -c 'bin/hdfs dfs -ls /'`
+`kubectl exec -it hdfs-namenode-default-0 -c namenode -- bash -c 'kdestroy && bin/hdfs dfs -ls /'`
```
nightkr: This paragraph doesn't really have the same oomph if we're kdestroying. Maybe we should change it to a `kubectl run` of some kind with a temporary pod instead?

rust/operator/src/container.rs
```diff
@@ -161,16 +157,22 @@ impl ContainerConfig {
...
         pb.add_container(main_container_config.main_container(
             hdfs,
             resolved_product_image,
             zk_config_map_name,
             env_overrides,
             merged_config,
-            object_name,
         )?);
 
+        // We need to share `/tmp` between all containers, e.g. for Kerberos ticket cache
```
nightkr: We shouldn't need to share the ticket cache as long as every container has access to the keytab itself (and container B relying on container A having populated the cache already is a great way to run into bugs when those tickets start expiring, for example).

rust/operator/src/container.rs
```diff
@@ -199,13 +203,40 @@ impl ContainerConfig {
...
-                    .command(vec!["/bin/bash".to_string(), "-c".to_string()])
+                    .command(Self::command())
                     .args(vec![formatdoc!(
                             r###"
+                            # We need get the kerberos ticket first, as other containers wait for the truststore to be available.
+                            # This way it is guaranteed that the ticket is already there when the truststore is there.
+                            export KRB5_CONFIG=/stackable/kerberos/krb5.conf
+                            {export_kerberos_real_env_var_command}
+                            echo "Getting ticket for {principal} from /stackable/kerberos/keytab"
+                            kinit "{principal}" -kt /stackable/kerberos/keytab
```
nightkr: At most this should only be relied on by other init containers, since these tickets expire (as mentioned above)...

rust/operator/src/container.rs
```diff
@@ -199,13 +203,40 @@ impl ContainerConfig {
...
+                        SecretOperatorVolumeSourceBuilder::new(
+                            &authentication_config.kerberos.secret_class,
+                        )
+                        .with_service_scope(hdfs.name_any())
+                        .with_kerberos_service_name(role.kerberos_service_name())
+                        .with_kerberos_service_name("HTTP")
+                        .build(),
+                    )
+                    .build(),
+            );
```
nightkr, line 212: :+1:

rust/operator/src/kerberos.rs
```diff
@@ -1,9 +1,21 @@
...
+use stackable_operator::commons::product_image_selection::ResolvedProductImage;
 
-use crate::config::{CoreSiteConfigBuilder, HdfsSiteConfigBuilder};
+use crate::{
+    config::{CoreSiteConfigBuilder, HdfsSiteConfigBuilder},
+    hdfs_controller::Error,
+};
+
+pub fn check_if_supported(resolved_product_image: &ResolvedProductImage) -> Result<(), Error> {
+    if resolved_product_image.product_version.starts_with("3.2.") {
```
nightkr, line 17: It feels hacky but I guess it's simple enough and works... Especially since we can drop it when we drop 3.2.x support.

rust/operator/src/kerberos.rs
```diff
@@ -22,80 +35,174 @@ impl HdfsSiteConfigBuilder {
...
         self
     }
+
+    fn add_wire_encryption_settings(&mut self) -> &mut Self {
+        self.add("hadoop.rpc.protection", "privacy");
+        self
+    }
 }
+
+// IMPORTANT: We only support Kerberos for HDFS >= 3.3.x
```
nightkr: Can we move this section up into the `check_if_supported` fn? Maybe with the full stack trace snipped out or moved into a gist/separate file.

tests/templates/kuttl/kerberos/21-unleash-the-chaosmonkey.yaml.j2
```diff
@@ -0,0 +1,49 @@
+# Tribute to https://github.com/Netflix/chaosmonkey
+
+# We need to force-delete the Pods, because IONOS is sometimes unable to delete the pod (it's stuck in Terminating for > 20 minutes)
```
nightkr, line 4: That's worrying... do we have access to IONOS' kubelet logs? Is this an IONOS-only issue?

tests/templates/kuttl/kerberos/21-unleash-the-chaosmonkey.yaml.j2
```diff
@@ -0,0 +1,49 @@
+# Tribute to https://github.com/Netflix/chaosmonkey
```
nightkr, line 1: Maybe add a link to #338 too, as a motivation for why this exists?

nightkr (commented):
Mostly LGTM in theory, a few nits and haven't tested it yet.

rust/operator/src/kerberos.rs
sbernauer: Exactly my thoughts :+1:

tests/templates/kuttl/kerberos/21-unleash-the-chaosmonkey.yaml.j2
sbernauer: Added in bfaafa4ffe28c96d9eb69c302b799f621bff64df

tests/templates/kuttl/kerberos/21-unleash-the-chaosmonkey.yaml.j2
sbernauer: Have to admit I'm not sure if we have access to the kubelet logs. Test passed on Azure however

rust/operator/src/kerberos.rs
sbernauer: I would prefer to keep the full stacktrace, just to be sure and have be prepared for customer questions.
Questions is where should I put it? A file `rust/operator/src/kerberos_hdfs_3.2_stacktrace.txt`?

docs/modules/hdfs/pages/usage-guide/security.adoc
sbernauer: Agreed. Would it be ok to get rid of this command entirely and rely on the Webinterface?

tests/templates/kuttl/kerberos/21-unleash-the-chaosmonkey.yaml.j2
sbernauer: Update: Support recommended using https://github.com/kvaps/kubectl-node-shell, which worked.
Sadly it only logs stuff during startup and than remains silent.
```
journalctl -u kubelet --since "48 hour ago"
-- Logs begin at Sat 2023-06-03 19:51:35 UTC, end at Tue 2023-06-06 12:01:05 UTC. --
-- No entries --
```
Anyway, even with force deletion tests have random timeouts on IONOS after we added the chaosmonkey. Everything is really slow

rust/operator/src/kerberos.rs
nightkr: Sure.

docs/modules/hdfs/pages/usage-guide/security.adoc
nightkr: I like demonstrating it with the CLI too, honestly. Maybe that's something to add to the stackablectl wishlist at some point...

rust/operator/src/container.rs
sbernauer: Implemented in 01d4a5a895bf8bddb4f18be66968e13af3e9e570

rust/operator/src/container.rs
sbernauer: Implemented in 01d4a5a895bf8bddb4f18be66968e13af3e9e570

docs/modules/hdfs/pages/usage-guide/security.adoc
nightkr: Do we still need to `kdestroy` now that we're back to CC-per-container? Actually, I think we could just override the CC path for that specific command...

rust/operator/src/kerberos.rs
sbernauer: 277c5503ffae6962acd3e55020dff9b0edc8194a

docs/modules/hdfs/pages/usage-guide/security.adoc
sbernauer: Current PR state has no kdestroy any more :+1:

sbernauer:
Should be ready to be merged :+1:

nightkr (approved):
Waiting for integration tests to complete, but LGTM otherwise.

We'll probably want to note this in the changelog though?

nightkr:
Tests passed.

sbernauer:
All that and I forgot the CHANGELOG :D
Thanks for your review and work + help on Kerberos overall!

I have 4 testclusters running since this morning, let's wait some time with merging to see if they survive

sbernauer:
All 4 clusters are still healthy, merging this

PR #402, opened 2023-09-29
fix: Don't default roleGroup replicas to zero when not specified

### LINKED ISSUE #398: Don't default roleGroup replicas to zero when not specified
### Affected version

0.0.0-dev

In https://github.com/stackabletech/hdfs-operator/blob/f55c8ff066116e657fd71049d555fbcd161611dc/rust/crd/src/lib.rs#L373-L392 we set the sts replicas to 0, when the user does not configure a `replicas` number on a rolegroup.
Instead we should change the fn to `pub fn role_group_replicas(&self, hdfs: &HdfsCluster, role_group: &str) -> Option<i32>` and pass the `Option<i32>` directly into the sts.
This way when the user does not specify replicas, the sts will have not set set replicas and k8s will default to 1 pod.
Users can than use [HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) to scale their Pods based on e.g. resource usage.
This is how other operators e.g. trino are doing it (the right way)

### Possible solution

_No response_

### Additional context

_No response_

### Environment

_No response_

None

sbernauer:
# Description

Fixes https://github.com/stackabletech/hdfs-operator/issues/398
Part of https://github.com/stackabletech/issues/issues/435

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated
- [x] Changelog updated
```

```[tasklist]
# Acceptance
- [x] Proper release label has been added
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,11 +16,16 @@ All notable changes to this project will be documented in this file.
 - Let secret-operator handle certificate conversion ([#392]).
 - `operator-rs` `0.44.0` -> `0.51.1` ([#381], [#394]).
 
+### Fixed
+
+- Don't default roleGroup replicas to zero when not specified ([#402]).
+
 [#378]: https://github.com/stackabletech/hdfs-operator/pull/378
 [#381]: https://github.com/stackabletech/hdfs-operator/pull/381
 [#384]: https://github.com/stackabletech/hdfs-operator/pull/384
 [#392]: https://github.com/stackabletech/hdfs-operator/pull/392
 [#394]: https://github.com/stackabletech/hdfs-operator/pull/394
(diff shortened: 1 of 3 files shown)


fhennig (commented):
LGTM, but what about an integration test for this? Might be a bit trivial but still :grimacing: 

Also, we should document the behaviour if the replicas are omitted.

sbernauer:
You might be interested in https://github.com/stackabletech/documentation/pull/461 :)
Regarding the integration test I see why we want it, but I think it this should be a unit test but well... we have problems with mocking k8s

fhennig (dismissed):

maltesander (approved):
LGTM!

PR #407, opened 2023-10-11
feat: Support graceful shutdown

### LINKED ISSUE #406: Support graceful shutdown
Part of https://github.com/stackabletech/issues/issues/357

Must
- [x] Code goes into `rust/operator-binary/src/operations/graceful_shutdown.rs` (if needed)
- [x] Update `docs/modules/<operator>/pages/usage-guide/operations/graceful-shutdown.adoc`
- [x] Functionality verified manually by restarting Pods
- [x] Graceful shutdown period must be configurable, probably on roleGroup `config` (see CRD changes tab). Implementation is copied from trino-operator and uses human-readable time.
- [x] Sane default for graceful shutdown period determined
- [x] smoke test tests default value

Should

- [x] Integration test added, e.g. something like chaos monkey [in hdfs](https://github.com/stackabletech/hdfs-operator/blob/main/tests/templates/kuttl/smoke/60-unleash-the-chaosmonkey.yaml.j2). This is probably the hardest part of the whole task, so it's a "Should"

sbernauer:
# Description

Part of https://github.com/stackabletech/issues/issues/357
Closes https://github.com/stackabletech/hdfs-operator/issues/406
Needs https://github.com/stackabletech/operator-rs/pull/681

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Documentation added or updated
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
+++ b/docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
@@ -7,6 +7,7 @@ You can configure the graceful shutdown as described in xref:concepts:operations
 As a default, JournalNodes have `15` minutes to terminate gracefully.
 
 In case they are requested to shut down, they will receive a `SIGTERM`, log the received signal as follows and shut down gracefully.
+https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java#L2004[This] is the relevant code as of version `3.3.4`.
 
 [source,text]
 ----
@@ -22,9 +23,12 @@ SHUTDOWN_MSG: Shutting down JournalNode at hdfs-journalnode-default-0/10.244.0.3
 As a default, NameNodes have `15` minutes to terminate gracefully.
 
 In case they are requested to shut down, they will receive a `SIGTERM`, log the received signal and shut down gracefully.
+https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/NameNode.java#L1080[This] is the relevant code as of version `3.3.4`.
+
 
 == DataNodes
 
 As a default, DataNodes have `15` minutes to terminate gracefully.
 
 In case they are requested to shut down, they will receive a `SIGTERM`, log the received signal and shut down gracefully.
+https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/qjournal/server/JournalNode.java#L272[This] is the relevant code as of version `3.3.4`.
--- a/docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
+++ b/docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
@@ -4,11 +4,11 @@ You can configure the graceful shutdown as described in xref:concepts:operations
 
 == JournalNodes
 
-As a default, JournalNodes have `15 minutes` to terminate gracefully.
+As a default, JournalNodes have `15 minutes` to shut down gracefully.
 
-The JournalNode process will always run as PID `1` and will get a `SIGTERM` once Kubernetes wants to terminate the Pod.
+The JournalNode process will always run as PID `1` and will receive a `SIGTERM` signal when Kubernetes wants to terminate the Pod.
 It will log the received signal as show in the log below and initiate a graceful shutdown.
-After the graceful shutdown timeout is passed and the process still didn't exit, Kubernetes will issue an `SIGKILL` to force-kill the process.
+After the graceful shutdown timeout runs out, and the process still didn't exit, Kubernetes will issue a `SIGKILL` signal.
 
 https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java#L2004[This] is the relevant code that gets executed in the JournalNodes as of HDFS version `3.3.4`.
 
@@ -23,14 +23,14 @@ SHUTDOWN_MSG: Shutting down JournalNode at hdfs-journalnode-default-0/10.244.0.3
 
 == NameNodes
 
-As a default, NameNodes have `15 minutes` to terminate gracefully.
-They go through the same mechanism as documented for the <<_journalnodes>> above.
+As a default, NameNodes have `15 minutes` to shut down gracefully.
+They use the same mechanism described above.
 
 https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/NameNode.java#L1080[This] is the relevant code that gets executed in the NameNodes as of HDFS version `3.3.4`.
 
 == DataNodes
 
-As a default, DataNodes have `30 minutes` to terminate gracefully.
-They go through the same mechanism as documented for the <<_journalnodes>> above.
+As a default, DataNodes have `30 minutes` to shut down gracefully.
+They use the same mechanism described above.
 
 https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/qjournal/server/JournalNode.java#L272[This] is the relevant code that gets executed in the DataNodes as of HDFS version `3.3.4`.
--- a/docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
+++ b/docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
@@ -1,5 +1,9 @@
 = Graceful shutdown
 
+Graceful shutdown refers to the managed, controlled shutdown of service instances in the manner intended by the software authors. Typically, an instance will receive a signal indicating the intent for the server to shut down, and it will initiate a controlled shutdown. This could include closing open file handles, updating the instance state in the cluster and emitting a message that the server is closing down. This contrasts with an uncontrolled shutdown where a process is terminated immediately and is unable to perform any of its normal shutdown activities.
+
+In the event that a service instance is unable to shut down in a reasonable amount of time, a timeout is set after which the process will be forcibly terminated to prevent a stuck server from remaining in the shutting down state indefinitely.
+
 You can configure the graceful shutdown as described in xref:concepts:operations/graceful_shutdown.adoc[].
 
 == JournalNodes
--- a/docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
+++ b/docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
@@ -6,8 +6,11 @@ You can configure the graceful shutdown as described in xref:concepts:operations
 
 As a default, JournalNodes have `15 minutes` to terminate gracefully.
 
-In case they are requested to shut down, they will receive a `SIGTERM`, log the received signal as follows and shut down gracefully.
-https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java#L2004[This] is the relevant code as of version `3.3.4`.
+The JournalNode process will always run as PID `1` and will get a `SIGTERM` once Kubernetes wants to terminate the Pod.
+It will log the received signal as show in the log below and initiate a graceful shutdown.
+After the graceful shutdown timeout is passed and the process still didn't exit, Kubernetes will issue an `SIGKILL` to force-kill the process.
+
+https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java#L2004[This] is the relevant code that gets executed in the JournalNodes as of HDFS version `3.3.4`.
 
 [source,text]
 ----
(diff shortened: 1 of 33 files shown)


sbernauer:
https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/101/

rust/operator/src/operations/graceful_shutdown.rs
```diff
@@ -0,0 +1,12 @@
+use stackable_hdfs_crd::MergedConfig;
+use stackable_operator::builder::PodBuilder;
+
+pub fn add_graceful_shutdown_config(
+    merged_config: &(dyn MergedConfig + Send + 'static),
+    pod_builder: &mut PodBuilder,
+) {
+    // This must be always set by the merge mechanism, as we provide a default value.
+    if let Some(graceful_shutdown_timeout) = merged_config.graceful_shutdown_timeout() {
+        pod_builder.termination_grace_period_seconds(graceful_shutdown_timeout.as_secs() as i64);
```
nightkr: Would it make sense to add a `PodBuilder::termination_grace_period` that takes our new sble `Duration`?

rust/crd/src/lib.rs
```diff
@@ -916,6 +927,7 @@ impl NameNodeConfigFragment {
             },
             logging: product_logging::spec::default_logging(),
             affinity: get_affinity(cluster_name, role),
+            graceful_shutdown_timeout: Some(Duration::from_minutes_unchecked(15)),
```
Techassi: I would like to see this defined as a constant (at the top of the file or in a separate `constants.rs` file (which would be even better)). Same goes for other definitions below.

rust/operator/src/operations/graceful_shutdown.rs
Techassi: This cast is dangerous. We basically half the capacity of a `u64` here. It's a bummer that `pod_builder.termination_grace_period_seconds()` takes an `i64`. We should maybe think about changing that (a negative amount of seconds doesn't even make sense).

Techassi (changes requested):

rust/operator/src/operations/graceful_shutdown.rs
nightkr:
```suggestion
        pod_builder.termination_grace_period_seconds(graceful_shutdown_timeout.as_secs().try_into().context(DurationTooLongSnafu { duration })?);
```

`foo as i64` will silently turn numbers above `9223372036854775808` negative. It's very much a corner-case here (anyone setting timers around over 200 billion years into the future has probably got bigger issues...), but I'd rather just blanket-ban the syntax (`#![deny(clippy::as_conversions)]` or even `#![deny(clippy::pedantic)]`...).

rust/operator/src/operations/graceful_shutdown.rs
```diff
@@ -0,0 +1,12 @@
+use stackable_hdfs_crd::MergedConfig;
+use stackable_operator::builder::PodBuilder;
+
+pub fn add_graceful_shutdown_config(
+    merged_config: &(dyn MergedConfig + Send + 'static),
+    pod_builder: &mut PodBuilder,
+) {
+    // This must be always set by the merge mechanism, as we provide a default value.
```
nightkr: That seems kind of incidental to me, this code would still be fully valid with no timeout set?

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
```diff
@@ -1,6 +1,33 @@
 = Graceful shutdown
 
-Graceful shutdown of HDFS nodes is either not supported by the product itself
-or we have not implemented it yet.
+You can configure the graceful shutdown as described in xref:concepts:operations/pod_placement.adoc[].
 
-Outstanding implementation work for the graceful shutdowns of all products where this functionality is relevant is tracked in https://github.com/stackabletech/issues/issues/357
+== JournalNodes
+
+As a default, JournalNodes have `15` minutes to terminate gracefully.
```
nightkr: Nit: the value is "15 minutes", they should (IMO) be formatted together.

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
```diff
@@ -1,6 +1,33 @@
...
+2023-10-10 13:37:41,525 ERROR server.JournalNode (LogAdapter.java:error(75)) - RECEIVED SIGNAL 15: SIGTERM
+2023-10-10 13:37:41,526 INFO  server.JournalNode (LogAdapter.java:info(51)) - SHUTDOWN_MSG:
+/************************************************************
+SHUTDOWN_MSG: Shutting down JournalNode at hdfs-journalnode-default-0/10.244.0.38
+************************************************************/
+----
+
+== NameNodes
+
+As a default, NameNodes have `15` minutes to terminate gracefully.
```
nightkr: Same here.

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
```diff
@@ -1,6 +1,33 @@
...
+
+In case they are requested to shut down, they will receive a `SIGTERM`, log the received signal and shut down gracefully.
+https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/NameNode.java#L1080[This] is the relevant code as of version `3.3.4`.
+
+== DataNodes
+
+As a default, DataNodes have `30` minutes to terminate gracefully.
```
nightkr: And here.

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
```diff
@@ -1,6 +1,33 @@
 = Graceful shutdown
 
-Graceful shutdown of HDFS nodes is either not supported by the product itself
-or we have not implemented it yet.
+You can configure the graceful shutdown as described in xref:concepts:operations/pod_placement.adoc[].
```
nightkr: What does graceful shutdown mean here, wait-until-all-outstanding-requests-are-serviced? Always wait out the timer before shutting down? Ensure "committed" writes are actually persisted? Ensure the standby namenode has taken over before shutting down the primary?

Why are we doing this, what would the consequences be of just SIGKILLing the daemon instead? Do the client libraries fail to retry if the node dies? Do we actually lose/corrupt data?

Those are all things I'd expect this documentation to answer.

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
nightkr: Also: where does this number come from? Do we have a reason to expect it to take this long?

rust/crd/src/lib.rs
```diff
@@ -841,6 +843,11 @@ pub struct NameNodeConfig {
     pub logging: Logging<NameNodeContainer>,
     #[fragment_attrs(serde(default))]
     pub affinity: StackableAffinity,
+    #[fragment_attrs(serde(default))]
+    #[fragment_attrs(schemars(
+        description = "Time period Pods have to gracefully shut down, e.g. `30m`, `1h` or `2d`. Consult the operator documentation for details."
+    ))]
```
nightkr: This looks like a fragmentization bug to me, we should probably copy over doccomments into the fragment field (and let schemars use that to generate the description).

rust/operator/src/operations/graceful_shutdown.rs
nightkr: FWIW the `i64` type comes from [k8s-openapi](https://docs.rs/k8s-openapi/0.20.0/k8s_openapi/api/core/v1/struct.PodSpec.html#structfield.termination_grace_period_seconds) so we can't really get rid of it entirely, but agreed that I'd rather hide the ugliness in `PodBuilder`.

Techassi: That sounds like a good idea. I have mentioned to same concerns about the `i64` you have above as well.

rust/crd/src/lib.rs
sbernauer: Moved in ce72ff82a3cb70e21ff2061be33295d8f5de5742

rust/operator/src/operations/graceful_shutdown.rs
sbernauer: +1 for `PodBuilder::termination_grace_period` :+1: It ended up the way it did, as `Duration` was not invented back than ;)
Created https://github.com/stackabletech/operator-rs/pull/672

Techassi: Yeah. I feel like its worth to change that upstream. Otherwise we just move the problem around (from the operator into the `PodBuilder`). Even the [k8s-openapi](https://docs.rs/k8s-openapi/0.20.0/k8s_openapi/api/core/v1/struct.PodSpec.html#structfield.termination_grace_period_seconds) documentation states:

> Value must be non-negative integer.

So the `i64` makes even less sense (if it made any sense in the first place).

Techassi: Relevant PR: https://github.com/stackabletech/operator-rs/pull/672

rust/operator/src/operations/graceful_shutdown.rs
sbernauer: Yeah, but I think it's a valid and important information that we will set a terminationGracePeriodSecond in all cases. Users can not disable it

rust/crd/src/lib.rs
sbernauer: Created https://github.com/stackabletech/operator-rs/pull/675

sbernauer: Merged it and started using it in dcb6bbb9aa18f5477f2d2326e8f333ba6bc173f5

rust/operator/src/operations/graceful_shutdown.rs
sbernauer: Now using the new operator-rs version with the `PodBuilder::termination_grace_period` fn

sbernauer: Now using the new operator-rs version with the `PodBuilder::termination_grace_period` fn
@Techassi can we resolve this thread?

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
sbernauer: To be honest it's a time period we did came up and considered to be a worst case, while being more on the pessimistic side

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
sbernauer: Sadly I don't know the answer to most of these and would need to research/read code. I'm happy to have our SME @lfrancke or @Jimvin to fill this gaps, but as discussed in Planning I wouldn't block this PR on that.

sbernauer:
@nightkr @Techassi all review comments should be addressed, please have a look!

nightkr (dismissed):
I'm okay with deferring the docs nits to a separate issue

rust/operator/src/operations/graceful_shutdown.rs
Techassi: Yes, we can resolve it for now.

Techassi (dismissed):

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
Jimvin: I've added a description of graceful shutdown to the docs as well as a mention of the timeout after which the process is forcibly terminated. That should add some more context and an explanation as to the meaning.

sbernauer:
@Jimvin thanks for the input!
I think this best belongs into the concepts guide, as this is shared between all operators. Created https://github.com/stackabletech/documentation/pull/473 with the content you added.
Would you be happy to merge that and revert your commit in this PR?

sbernauer:
Had a chat on Slack and agreed :)

rust/operator/src/operations/graceful_shutdown.rs
sbernauer: Can this be resolved? :)

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
sbernauer: Can this be resolved? :)

Techassi (dismissed):

sbernauer:
https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/104/

nightkr (dismissed):
No changes since the last review… Agreed that the docs-as-written-before-getting-reverted belong in the platform docs. Disagree that those docs adequately cover why this feature is needed for any one specific product (including HDFS).

sbernauer:
Agreed with @nightkr that product-specific docs on the product internals would be a great follow-up :+1:

sbernauer:
Thanks for all the reviews, also within in the sub PRs e.g. in operator-rs!

rust/operator/src/hdfs_controller.rs
```diff
@@ -171,7 +171,7 @@ pub enum Error {
         source: stackable_operator::product_config::writer::PropertiesWriterError,
         rolegroup: String,
     },
-    #[snafu(display("failed to configure graceful shutdown"), context(false))]
+    #[snafu(display("failed to configure graceful shutdown"))]
```
Techassi: Add newline.

Techassi (changes requested):
Final nit :)

rust/operator/src/hdfs_controller.rs
sbernauer: done

Techassi (dismissed):

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
```diff
@@ -1,6 +1,36 @@
...
-Graceful shutdown of HDFS nodes is either not supported by the product itself
-or we have not implemented it yet.
+You can configure the graceful shutdown as described in xref:concepts:operations/graceful_shutdown.adoc[].
 
-Outstanding implementation work for the graceful shutdowns of all products where this functionality is relevant is tracked in https://github.com/stackabletech/issues/issues/357
+== JournalNodes
+
+As a default, JournalNodes have `15 minutes` to shut down gracefully.
+
+The JournalNode process will always run as PID `1` and will receive a `SIGTERM` signal when Kubernetes wants to terminate the Pod.
```
razvan, line 9: Is it really relevant  that the process runs with PID 1 ? This is not true on OpenShift for example.

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
```diff
@@ -1,6 +1,36 @@
...
+It will log the received signal as show in the log below and initiate a graceful shutdown.
+After the graceful shutdown timeout runs out, and the process still didn't exit, Kubernetes will issue a `SIGKILL` signal.
+
+https://github.com/apache/hadoop/blob/a585a73c3e02ac62350c136643a5e7f6095a3dbb/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/DataNode.java#L2004[This] is the relevant code that gets executed in the JournalNodes as of HDFS version `3.3.4`.
```
razvan: This points to the DataNode and not JournalNode source. Typo ?

sbernauer: Thx, fixed

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
sbernauer: I think our stuff will always run as pid 1 as we don't use `shareProcessNamespace: true`.
I only have your "razvan" OS cluster at hand, and picked a random Pod, OS seems to behave normally
```
kubectl -n openshift-monitoring exec -it alertmanager-main-0 -- bash -c "ls /proc && cat /proc/1/stat"
1   acpi       cgroups   cpuinfo  diskstats  execdomains  fs          ioports   kcore      kmsg         kpageflags  mdstat   modules  net           sched_debug  self      stat   sysrq-trigger  timer_list  version      zoneinfo
60  buddyinfo  cmdline   crypto   dma        fb           interrupts  irq       key-users  kpagecgroup  loadavg     meminfo  mounts   pagetypeinfo  schedstat    slabinfo  swaps  sysvipc        tty         vmallocinfo
66  bus        consoles  devices  driver     filesystems  iomem       kallsyms  keys       kpagecount   locks       misc     mtrr     partitions    scsi         softirqs  sys    thread-self    uptime      vmstat
1 (alertmanager) S 0 1 1 0 -1 1077952768 3591272 0 8 0 136784 67329 0 0 20 0 11 0 5023 1526284288 14384 18446744073709551615 4194304 27613560 140722030886976 0 0 0 1002060288 0 2143420159 0 0 0 17 2 0 0 1 0 0 29711832 30179416 52162560 140722030890179 140722030890741 140722030890741 140722030895078 0
```
We can see the main process is running as pid 1

razvan: My bad. You are right. I was thinking of user id not process id.

sbernauer: Puh, had me sweating for a moment ^^

docs/modules/hdfs/pages/usage-guide/operations/graceful-shutdown.adoc
```diff
@@ -1,6 +1,37 @@
...
-or we have not implemented it yet.
+You can configure the graceful shutdown as described in xref:concepts:operations/graceful_shutdown.adoc[].
 
-Outstanding implementation work for the graceful shutdowns of all products where this functionality is relevant is tracked in https://github.com/stackabletech/issues/issues/357
+== JournalNodes
+
+As a default, JournalNodes have `15 minutes` to shut down gracefully.
+
+The JournalNode process will always run as PID `1` and will receive a `SIGTERM` signal when Kubernetes wants to terminate the Pod.
+It will log the received signal as show in the log below and initiate a graceful shutdown.
```
siegfriedweber:
```suggestion
It will log the received signal as shown in the log below and initiate a graceful shutdown.
```

siegfriedweber (changes requested):

siegfriedweber (approved):
LGTM

sbernauer:
Final test run: https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/107/

sbernauer:
Tests passed :rocket:

PR #426, opened 2023-11-16
Do not test for storage status

razvan:
Not all storage providers honor the storage sizes. GKE and CRC always provision at least 10Gi and 30Gi respectively.


dervoeti (approved):
LGTM since we still assert that status.phase == "Bound"

razvan:
thanks!

Issue #441, opened 2023-12-18
Roll back suggested HDFS to 3.3.4


nightkr:
#440 means that we're generating broken clusters, which is... not great.

Issue #440, opened 2023-12-18
Secondary NameNode bootstrap gets stuck on HDFS 3.3.6


nightkr:
### Affected version

23.11.0

### Current and expected behavior

Typically, one namenode does an initial format, and then another tries to bootstrap from the first.

However, for some reason the secondary seems to get stuck bootstrapping in some clusters, never letting the cluster initialize fully.

3.3.4 does not seem to show the same issue.

### Environment

```
Client Version: v1.27.4
Kustomize Version: v5.0.1
Server Version: v1.27.3
```

Deployed using ctlptl 0.8.22, kind 0.20.0.

### Would you like to work on fixing this bug?

None

nightkr:
Another maybe-related symptom of this:

After starting an old cluster back up neither namenode was able to become Active. They'd wait for their blocks and try to transition into active, but would get stuck halfway, eventually giving up and being restarted as standby, waiting for the cycle to repeat again.

fhennig:
@lfrancke this ticket hasn't moved since March 6, what is needed to move this along?

lfrancke:
I need to get to it at some point. This is mostly a reminder at this point to not forget about it.

PR #450, opened 2024-01-08
Listener support/external access

### LINKED ISSUE #174: Allow external access to the HDFS cluster

nightkr:
# Description

Fixes #174, supersedes #175.

```[tasklist]
# Tasks before R4R
- [x] Support external access
- [x] Support Kerberos
- [ ] https://github.com/stackabletech/docker-images/pull/506
- [x] Move listenerclass into role config? https://stackable-workspace.slack.com/archives/C02FZ581UCD/p1704714818413279
- [ ] Add proper listener watch for discovery
- [x] Docs
- [ ] Tests
```

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


soenkeliebau:
Should we close #175 to avoid confusion?

nightkr:
Sounds good yeah.

nightkr:
Watching is blocked on https://github.com/stackabletech/listener-operator/issues/138, but I wouldn't block on that since most expected changes here would be during startup/scale events (which would ping the operator anyway).

We'd still want to pull that out into an issue here and fix eventually, obviously.

nightkr:
The current testing approach is blocked on https://github.com/kudobuilder/kuttl/issues/322. Are we okay to punt on testing external access as long as the internal access tests still function?

tests/test-definition.yaml
```diff
@@ -3,8 +3,11 @@ dimensions:
   - name: hadoop
     values:
       - 3.2.4
-      - 3.3.6
+      - 3.3.4
```
nightkr, line 6: #440, #441

docs/modules/hdfs/pages/usage-guide/listenerclass.adoc
```diff
@@ -1,15 +1,18 @@
 = Service exposition with ListenerClasses
 
-The Operator deploys a service called `<name>-<role>-<rolegroup>` (where `<name>` is the name of the HdfsCluster, `<role>` is the role and `<rolegroup>` the name of the role group) through which the different HDFS processes can be accessed. Unlike many other Stackable Operators, the HDFS Operator does not deploy role-level Services.
+The Operator deploys a Listener for each DataNode and NameMode pod. They both default to only being accessible from within the Kubernetes cluster, but this can be changed by changing `.spec.{data,name}Nodes.config.listenerClass`.
```
siegfriedweber: * There should be a link to the Listener or ListenerClass documentation.
* NameMode -> NameNode

```suggestion
The operator deploys a xref:listener-operator:listener.adoc[Listener] for each DataNode and NameNode pod. They both default to only being accessible from within the Kubernetes cluster, but this can be changed by setting `.spec.{data,name}Nodes.config.listenerClass`.
```

rust/operator-binary/src/container.rs
```diff
@@ -291,24 +292,50 @@ impl ContainerConfig {
...
+                    .volume_claim_template
+                    .unwrap();
+                    PersistentVolumeClaim {
+                        metadata: ObjectMeta {
+                            name: Some(LISTENER_VOLUME_NAME.to_string()),
+                            ..listener.metadata.unwrap()
+                        },
+                        spec: Some(listener.spec),
+                        ..Default::default()
+                    }
```
siegfriedweber: The new `ListenerOperatorVolumeSourceBuilder::build_pvc` from operator-rs is probably meant to be used here.

rust/operator-binary/src/container.rs
```diff
@@ -769,7 +802,7 @@ wait_for_termination $!
         let mut volumes = vec![];
 
         let container_log_config = match self {
```
siegfriedweber: **nit:** The following code block should return the container log config (`let container_log_config = ...`). The existing side effect that also the `volumes` variable is populated with a log volume, is inelegant, but at least logging related. I would not add code to this code block, which is not logging related.

rust/operator-binary/src/container.rs
```diff
@@ -787,6 +820,23 @@ wait_for_termination $!
...
+                    if let Some(listener_class) = merged_config.listener_class() {
+                        volumes.push(
+                            VolumeBuilder::new(LISTENER_VOLUME_NAME)
+                                .ephemeral(
+                                    ListenerOperatorVolumeSourceBuilder::new(
+                                        &ListenerReference::ListenerClass(
+                                            listener_class.to_string(),
+                                        ),
+                                    )
+                                    .build(),
```
siegfriedweber: After upgrading to the latest operator-rs version:

```suggestion
                                    .build_ephemeral(),
```

tests/templates/kuttl/external-access/00-limit-range.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v1
+kind: LimitRange
+metadata:
+  name: limit-request-ratio
```
siegfriedweber: In my opinion, it is sufficient to have the LimitRange in the smoke test.

deploy/helm/hdfs-operator/crds/crds.yaml
```diff
@@ -57,20 +57,6 @@ spec:
...
-
-                        * cluster-internal: Use a ClusterIP service
-
-                        * external-unstable: Use a NodePort service
-
-                        This is a temporary solution with the goal to keep yaml manifests forward compatible. In the future, this setting will control which [ListenerClass](https://docs.stackable.tech/home/nightly/listener-operator/listenerclass.html) will be used to expose the service, and ListenerClass names will stay the same, allowing for a non-breaking change.
-                      enum:
-                        - cluster-internal
-                        - external-unstable
-                      type: string
```
siegfriedweber, line 73: Are the CRD changes approved?

siegfriedweber (changes requested):

deploy/helm/hdfs-operator/crds/crds.yaml
nightkr: They were discussed in https://stackable-workspace.slack.com/archives/C02FZ581UCD/p1704714818413279, and this _re_aligns with the original ADR.

I'm fine adding it to the arch discussion list for next week.

rust/operator-binary/src/container.rs
nightkr: You're right, I shouldn't have tunnel visioned on "OH THIS IS THE BIG MATCH".

nightkr: 3169a4f9624a4a6779523fd73c276b1a8b31a086

rust/operator-binary/src/container.rs
nightkr: Exactly.

nightkr: Upgrading to op-rs 0.62 seems to bring with it a bunch of other breaking changes that I would consider out of scope for this PR.

tests/templates/kuttl/external-access/00-limit-range.yaml
nightkr: 86a32296f617643e6649cfa743ef121103e7ecce

rust/operator-binary/src/container.rs
siegfriedweber: You are right. We already have a pull request for upgrading operator-rs, so I moved this change request there: https://github.com/stackabletech/hdfs-operator/pull/454#issuecomment-1905472546

deploy/helm/hdfs-operator/crds/crds.yaml
siegfriedweber: approved

siegfriedweber (approved):

PR #424, opened 2023-11-09
fix: Include hdfs principal names in discovery ConfigMap

sbernauer:
# Description

Fixes https://github.com/stackabletech/hdfs-operator/issues/419
Fixed while working on https://github.com/stackabletech/hbase-operator/issues/285.
Strangely HBase works without the principals being configured...

## Definition of Done Checklist

- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
- Please make sure all these things are done and tick the boxes
 
```[tasklist]
# Author
- [ ] Changes are OpenShift compatible
- [ ] CRD changes approved
- [ ] Helm chart can be installed and deployed operator works
- [ ] Integration tests passed (for non trivial changes)
- [ ] Changes need to be "offline" compatible
```

```[tasklist]
# Reviewer
- [ ] Code contains useful comments
- [ ] (Integration-)Test cases added
- [ ] Documentation added or updated
- [ ] Changelog updated
- [ ] Cargo.toml only contains references to git tags (not specific commits or branches)
```

```[tasklist]
# Acceptance
- [ ] Feature Tracker has been updated
- [ ] Proper release label has been added
```

PR #451, opened 2024-01-09
fix: Include hdfs principal names in discovery ConfigMap

### LINKED ISSUE #419: Discovery configmap incomplete when Kerberos is enabled
### Affected version

23.7.0

When Kerberos is enabled on HDFS a number of properties must be added to core-site.xml to enable clients to determine the correct kerberos principal for the services. Without this the client will be unable to determine which principal to use when requesting a Kerberos ticket and you see the following error in the client logs:

```
java.lang.IllegalArgumentException: Failed to specify server's Kerberos principal name, while invoking ClientNamenodeProtocolTranslatorPB.getListing over poc-hdfs-namenode-default-0.poc-hdfs-namenode-default.default.svc.cluster.local/10.244.3.19:8020
```

### Possible solution

The following properties should be added to core-site.xml in the  HDFS discovery configMap when Kerberos is enabled:

- dfs.datanode.kerberos.principal
- dfs.namenode.kerberos.principal
- dfs.web.authentication.kerberos.principal

### Additional context

_No response_

### Environment

_No response_

None

sbernauer:
# Description

Replaces https://github.com/stackabletech/hdfs-operator/pull/424
Fixes https://github.com/stackabletech/hdfs-operator/issues/419

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
- [x] Proper release label has been added
```


docs/modules/hdfs/pages/reference/discovery.adoc
```diff
@@ -37,4 +37,13 @@ The ConfigMap data values are formatted as Hadoop XML files which allows simple
 Contains the `fs.DefaultFS` which defaults to `hdfs://{clusterName}/`.
 
 `hdfs-site.xml`::
-Contains the `dfs.namenode.*` properties for `rpc` and `http` addresses for the `namenodes` as well as the `dfs.nameservices` property which defaults to `hdfs://{clusterName}/`.
\ No newline at end of file
+Contains the `dfs.namenode.*` properties for `rpc` and `http` addresses for the `namenodes` as well as the `dfs.nameservices` property which defaults to `hdfs://{clusterName}/`.
+
+=== Kerberos
+In case Kerberos is enabled according the xref:usage-guide/security.adoc[security documentation] the discovery will also include the information that clients must authenticate themselves using Kerberos.
```
siegfriedweber:
```suggestion
In case Kerberos is enabled according to the xref:usage-guide/security.adoc[security documentation], the discovery ConfigMap also includes the information that clients must authenticate themselves using Kerberos.
```

docs/modules/hdfs/pages/reference/discovery.adoc
```diff
@@ -37,4 +37,13 @@ The ConfigMap data values are formatted as Hadoop XML files which allows simple
...
+
+Some Kerberos-related configuration settings require the environment variable `KERBEROS_REALM` to be set (e.g. using `export KERBEROS_REALM=$(grep -oP 'default_realm = \K.*' /stackable/kerberos/krb5.conf)`).
+When you want to use the discovery ConfigMap outside of Stackable services, you need to provide this environment variable!
```
siegfriedweber:
```suggestion
If you want to use the discovery ConfigMap outside Stackable services, you need to provide this environment variable.
```

docs/modules/hdfs/pages/reference/discovery.adoc
```diff
@@ -37,4 +37,13 @@ The ConfigMap data values are formatted as Hadoop XML files which allows simple
...
+As an alternative you can text-replace `${env.KERBEROS_REALM}` with your actual realm (e.g. by using `sed -i -e 's/${{env.KERBEROS_REALM}}/'"$KERBEROS_REALM/g" hbase-site.xml`).
```
siegfriedweber:
```suggestion
As an alternative you can substitute `${env.KERBEROS_REALM}` with your actual realm (e.g. by using `sed -i -e 's/${{env.KERBEROS_REALM}}/'"$KERBEROS_REALM/g" core-site.xml`).
```

docs/modules/hdfs/pages/reference/discovery.adoc
```diff
@@ -37,4 +37,13 @@ The ConfigMap data values are formatted as Hadoop XML files which allows simple
...
+
+One example would be the property `dfs.namenode.kerberos.principal` being set to `nn/hdfs.kuttl-test-daring-mammal.svc.cluster.local@${env.KERBEROS_REALM}`.
```
siegfriedweber:
```suggestion
One example would be the property `dfs.namenode.kerberos.principal` being set to `nn/hdfs.default.svc.cluster.local@${env.KERBEROS_REALM}`.
```

rust/operator-binary/src/discovery.rs
```diff
@@ -21,12 +21,16 @@ pub fn build_discovery_configmap(
     controller: &str,
     namenode_podrefs: &[HdfsPodRef],
     resolved_product_image: &ResolvedProductImage,
-) -> OperatorResult<ConfigMap> {
+) -> Result<ConfigMap, crate::hdfs_controller::Error> {
```
siegfriedweber, line 24: **nit:** Snafu proposes to have an error type per module (https://docs.rs/snafu/latest/snafu/guide/philosophy/index.html#many-error-types).

rust/operator-binary/src/kerberos.rs
```diff
@@ -52,29 +55,14 @@ impl HdfsSiteConfigBuilder {
 }
 
 impl CoreSiteConfigBuilder {
-    pub fn security_config(
-        &mut self,
-        hdfs: &HdfsCluster,
-        hdfs_name: &str,
-        hdfs_namespace: &str,
-    ) -> &mut Self {
+    pub fn security_config(&mut self, hdfs: &HdfsCluster) -> Result<&mut Self, Error> {
```
siegfriedweber, line 58: **nit:** The former decision to pass validated parameters, was not so bad.

siegfriedweber (changes requested):

rust/operator-binary/src/discovery.rs
sbernauer: Yeah we currently have a mess, I want that as well. We need to address that at some point of time, I just fear this is out of scope for this PR

rust/operator-binary/src/kerberos.rs
sbernauer: Well I basically created this PR because of https://github.com/stackabletech/hdfs-operator/pull/424#discussion_r1396878200.
I think I'm with @maltesander  here, as I needed to pass them around everywhere... But not a strong feeling

siegfriedweber (approved):
LGTM

sbernauer:
Thanks for the review!

rust/operator-binary/src/discovery.rs
NickLarsenNZ: FYI, #454 brings in some Snafu enums

rust/operator-binary/src/discovery.rs
```diff
@@ -21,12 +21,16 @@ pub fn build_discovery_configmap(
...
     ConfigMapBuilder::new()
         .metadata(
             ObjectMetaBuilder::new()
                 .name_and_namespace(hdfs)
-                .ownerreference_from_resource(hdfs, None, Some(true))?
+                .ownerreference_from_resource(hdfs, None, Some(true))
+                .map_err(|err| Error::ObjectMissingMetadataForOwnerRef {
```
NickLarsenNZ, line 30: @sbernauer I believe we can use .with_context instead of .map_err here, eg:

```rust
        .with_context(|_| ObjectMissingMetadataForOwnerRefSnafu {
            obj_ref: ObjectRef::from_obj(hdfs),
        })
```

_the source gets set inside with_error, so doesn't need specifying here_

NickLarsenNZ: Although, it looks like we can simply do:

```rust
        .context(ObjectMissingMetadataForOwnerRefSnafu {
            obj_ref: ObjectRef::from_obj(hdfs),
        })?
```

So I'm unsure when to use with_context(|e: Error| ...) vs context(...) when you don't need access to the Error

Techassi: Yes, if the field is called `source`, the source error will automatically be picked up. So using `context()` instead of `with_context()` is recommended.

rust/operator-binary/src/discovery.rs
sbernauer: Awesome, thanks!

rust/operator-binary/src/discovery.rs
sbernauer: Will give that a try next time :+1:

NickLarsenNZ: done in #454

NickLarsenNZ: More on this while talking to @nightkr...

Just curious, I see a lot of `with_context` used over the place. Often the error variable is being ignored, so are these equivalent?

```rust
                    .with_context(|| MissingRoleSnafu {
                        role: HdfsRole::NameNode.to_string(),
                    })?;
```

```rust
                    .context(MissingRoleSnafu {
                        role: HdfsRole::NameNode.to_string(),
                    })?;
```

> @nightkr: the difference is that `.context()` is always evaluated, while `.with_context()` is only evaluated if there is an error to add context to
> it doesn't really matter for trivial context selectors, but it lets us avoid `to_string()`/`clone()` calls in the happy case
> but arguably in that case, the error should just take a HdfsRole instead of a `String`, resolving that neatly anyway

So just to make sure I understand, if `HdfsRole` implemented `Into<String>`, then we could do this at no expense?

```rust
                    .context(MissingRoleSnafu {
                        role: HdfsRole::NameNode, // into::<String>() will be evaluated if the error actually occurs
                    })?;
```

> @nightkr: yes

nightkr: The difference is that `with_context()` will short-circuit, and only evaluate the ctx in an actual error case, while `context()` will always be evaluated. `with_context()` should be preferred if the ctx is non-trivial (like calling `Clone::clone()` or `Display::to_string()`).

PR #454, opened 2024-01-11
chore: bump operator-rs to 0.62.0

NickLarsenNZ:
# Description

- **BREAKING** Removed legacy node selector on roleGroups.
- Bump operator-rs to 0.62.0 and Rust toolchain to 1.75.0
- Use new label builders.
- Code style changes

```[tasklist]
# Author
- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/style-guide).
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/style-guide).
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


sbernauer:
Sorry for bringing the bad news but there are some merge conflicts with https://github.com/stackabletech/hdfs-operator/pull/457

NickLarsenNZ:
> Sorry for bringing the bad news but there are some merge conflicts with #457

Story of my day @sbernauer... will fix and request approval

siegfriedweber:
Please use `ListenerOperatorVolumeSourceBuilder::build_pvc` in `container::ContainerConfig::volume_claim_templates` when #450 is merged.

nightkr:
Oops, missed this and ended up doing it in parallel.. dunno if there's anything relevant to grab from it but https://gist.github.com/nightkr/da4eb77f001dfb59a85bc43cbff39634

NickLarsenNZ:
> Oops, missed this and ended up doing it in parallel.. dunno if there's anything relevant to grab from it but https://gist.github.com/nightkr/da4eb77f001dfb59a85bc43cbff39634

@nightkr maybe in a new PR? I'd like to get this in now so others can build off of it as the merge conflicts are getting tricky to resolve.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - `operator-rs` `0.56.1` -> `0.57.0` ([#433]).
```
nightkr: This note is outdated, should either be removed or updated IMO.

rust/crd/src/lib.rs
```diff
@@ -54,16 +51,24 @@ pub mod constants;
...
     MissingRole { role: String },
+
     #[snafu(display("Missing role group [{role_group}] for role [{role}]"))]
     MissingRoleGroup { role: String, role_group: String },
+
     #[snafu(display("fragment validation failure"))]
     FragmentValidationFailure { source: ValidationError },
+
+    #[snafu(display("failed to build label"))]
+    BuildLabel { source: LabelError },
```
nightkr: Which set of labels, in which context?

rust/operator-binary/src/hdfs_controller.rs
```diff
@@ -233,8 +243,8 @@ pub async fn reconcile_hdfs(hdfs: Arc<HdfsCluster>, ctx: Arc<Ctx>) -> HdfsOperat
         .spec
         .image
         .resolve(DOCKER_IMAGE_BASE_NAME, crate::built_info::CARGO_PKG_VERSION);
-    if hdfs.has_kerberos_enabled() {
-        kerberos::check_if_supported(&resolved_product_image)?;
+    if hdfs.has_kerberos_enabled() && kerberos::check_if_supported(&resolved_product_image) {
```
nightkr: I would assume that `check_if_supported` returns `true` on success (though `is_supported` would arguably be clearer).

rust/operator-binary/src/kerberos.rs
```diff
@@ -7,21 +8,25 @@ use stackable_operator::{
...
+    ObjectHasNoNamespace {
+        source: stackable_hdfs_crd::Error,
+        obj_ref: ObjectRef<HdfsCluster>,
+    },
+}
+
+/// We only support Kerberos for HDFS >= 3.3.x
+/// With HDFS 3.2.2 we got weird errors, which *might* be caused by DNS lookup issues
+/// The Stacktrace is documented in rust/operator/src/kerberos_hdfs_3.2_stacktrace.txt
+pub fn check_if_supported(resolved_product_image: &ResolvedProductImage) -> bool {
```
nightkr: Alternatively..

```suggestion
#[derive(Debug, Snafu)]
struct KerberosNotSupported;
pub fn require_support(resolved_product_image: &ResolvedProductImage) -> Result<(), KerberosNotSupported> {
```

rust/operator-binary/src/kerberos.rs
```diff
@@ -147,11 +152,11 @@ impl CoreSiteConfigBuilder {
 /// ```
 ///
 /// After we have switched to using the following principals everything worked without problems
-fn principal_host_part(hdfs: &HdfsCluster) -> Result<String, Error> {
+fn principal_host_part(hdfs: &HdfsCluster) -> Result<String> {
     let hdfs_name = hdfs.name_any();
     let hdfs_namespace = hdfs
         .namespace_or_error()
-        .map_err(|_| Error::ObjectHasNoNamespace {
+        .context(ObjectHasNoNamespaceSnafu {
```
nightkr:
```suggestion
        .with_context(|_| ObjectHasNoNamespaceSnafu {
```

rust/operator-binary/src/pod_svc_controller.rs
```diff
@@ -47,21 +58,28 @@ pub struct Ctx {
...
-    let pod_labels = pod
-        .metadata
-        .labels
-        .as_ref()
-        .with_context(|| PodHasNoLabelsSnafu { name: name.clone() })?;
+    let pod_labels = Labels::try_from(
+        pod.metadata
+            .labels
+            .as_ref()
+            .with_context(|| PodHasNoLabelsSnafu { name: name.clone() })?,
```
nightkr:
```suggestion
            .as_ref_or_else(BTreeMap::default),
```

Then again, this is all going away soon anyway.

rust/operator-binary/src/pod_svc_controller.rs
```diff
@@ -47,21 +58,28 @@ pub struct Ctx {
...
 
     let recommended_labels_from_pod = pod_labels
         .iter()
-        .filter(|(key, _)| key.starts_with(APP_KUBERNETES_LABEL_BASE))
-        .map(|(key, value)| (key.clone(), value.clone()))
+        .filter(|&label| {
+            label
+                .key()
+                .prefix()
+                .is_some_and(|prefix| *prefix == APP_KUBERNETES_LABEL_BASE)
```
nightkr:
```suggestion
                == Some(APP_KUBERNETES_LABEL_BASE)
```

CHANGELOG.md
NickLarsenNZ: Happy to remove it (I heard somewhere that we aren't recording operator-rs bumps in the changelogs anymore).

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: Nice, yeah we had wondered why it would error out rather than just _not have any labels_.

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: Unfortunately not:

```
mismatched types
expected reference `&KeyPrefix`
   found reference `&'static str`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2Fhome%2Fnick%2Fdev%2Fgithub.com%2Fstackabletech%2Fhdfs-operator%2Frust%2Foperator-binary%2Fsrc%2Fpod_svc_controller.rs)
```

And using `KeyPrefix::from_str(...)` or ``KeyPrefix::try_from(...)` causes a result.

From memory I had already tried doing something like `impl<T> PartialEq<T> for Option<&KeyPrefix> where T: AsRef<str>`, but of course, The trait and the enum are foreign.

Any suggestions, or leave as is?

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: Hmm, I can't seem to get that working. Maybe I need to jump on a call with you.

rust/operator-binary/src/kerberos.rs
NickLarsenNZ: I turned it into a bool to make the body simpler, and so that we didn't need duplicate Snafus up the call chain.

rust/crd/src/lib.rs
NickLarsenNZ: You're gonna hate this, but it's everywhere (copy/pasted from the first operator being upgraded): https://github.com/search?q=org%3Astackabletech%20BuildLabelSnafu&type=code

Which do you prefer in this case:

- [ ] Consistency with the other operators, or
- [x] Remove ambiguity for this operator (and eventually do the same for other operators)?

CHANGELOG.md
NickLarsenNZ: Removed in 6bf0af7e90a49c6f5b8f642608202ecdf7da2a00

rust/operator-binary/src/hdfs_controller.rs
NickLarsenNZ: Renamed in 8b25f33ab422c3fd908dc1de226e6896f185e4d3

rust/operator-binary/src/kerberos.rs
NickLarsenNZ: Renamed in 8b25f33ab422c3fd908dc1de226e6896f185e4d3

Still a bool, see https://github.com/stackabletech/hdfs-operator/pull/454#discussion_r1462951107

rust/operator-binary/src/pod_svc_controller.rs
nightkr: What error do you get?

rust/operator-binary/src/pod_svc_controller.rs
nightkr: `KeyPrefix` is `Deref<Target = String>`, which is in turn `Deref<Target = str>`, so try `.prefix().as_deref().as_deref() == ...`. `&String` is kind of useless, ideally `KeyPrefix` would be changed to be `Deref<Target = str>` instead (which would let us get rid of one `.as_deref()`).

rust/operator-binary/src/kerberos.rs
nightkr: That's still inverted, returning `true` if the version is *not* supported.

rust/crd/src/lib.rs
nightkr: Scout's rule, IMO. We don't have to fix everywhere right now, but we should still try to improve the things we're working on.

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: Hmm, none of these are working:

```rust
        .filter(|&label| {
            label.key().prefix().as_deref().as_deref() == Some(APP_KUBERNETES_LABEL_BASE)
        })
```

```
mismatched types
expected reference `&KeyPrefix`
   found reference `&'static str`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2Fhome%2Fnick%2Fdev%2Fgithub.com%2Fstackabletech%2Fhdfs-operator%2Frust%2Foperator-binary%2Fsrc%2Fpod_svc_controller.rs)
```


But honestly, I don't think that is much tidier than how it was:

```rust
                .is_some_and(|prefix| *prefix == APP_KUBERNETES_LABEL_BASE)
```

nightkr: Aight, fair.

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: If I `use stackable_operator::utils::OptionExt;`

```
no method named `as_ref_or_else` found for enum `std::option::Option` in the current scope
items from traits can only be used if the trait is in scope
```


and if I `use stackable_operator::utils::OptionExt as _;`

```
the trait bound `stackable_operator::kvp::Labels: From<Cow<'_, BTreeMap<std::string::String, std::string::String>>>` is not satisfied
the following other types implement trait `TryFrom<T>`:
  <stackable_operator::kvp::Labels as TryFrom<BTreeMap<K, V>>>
  <stackable_operator::kvp::Labels as TryFrom<[(K, V); N]>>
  <stackable_operator::kvp::Labels as TryFrom<&BTreeMap<K, V>>>
required for `Cow<'_, BTreeMap<std::string::String, std::string::String>>` to implement `Into<stackable_operator::kvp::Labels>`
required for `stackable_operator::kvp::Labels` to implement `TryFrom<Cow<'_, BTreeMap<std::string::String, std::string::String>>>`
```

and the rest of the code starts showing a lot of errors because of inferred types. Probably easier to share my screen.

rust/operator-binary/src/kerberos.rs
NickLarsenNZ: Ohhh good point, I missed that part.

Resolved in e7160ad97d7beb9d96e057a07b3e0d87e993805e

rust/operator-binary/src/hdfs_controller.rs
NickLarsenNZ: Fixed in e7160ad97d7beb9d96e057a07b3e0d87e993805e

rust/crd/src/lib.rs
NickLarsenNZ: So this error variant is used a number of times, eg: https://github.com/stackabletech/hdfs-operator/blob/7160e88a832ff7588cc8cef24feef11000c2ea05/rust/crd/src/lib.rs#L475-L491

I don't think those should be converted to individual error variants, so perhaps some context could be added?

How about:

```rust
    #[snafu(display("failed to parse group label: {key}={value}"))]
    ParseGroupLabel { 
        source: LabelError,
        key: String,
        value: String,
    },
```

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: @Techassi is looking into the `Deref<Target = str>` in https://github.com/stackabletech/operator-rs/pull/725, so could be something to come back to

Techassi: I implemented the change from `String` to `str` as the `Deref` target in https://github.com/stackabletech/operator-rs/pull/725. While adding a few tests for this feature, I encountered a rather peculiar issue:

```rust
let label = Label::try_from((key, "zookeeper")).unwrap();

// Works as expected
let is_valid = label
    .key()
    .prefix()
    .is_some_and(|prefix| *prefix == "app.kubernetes.io");

// Doesn't work as expected
let is_valid = label.key().prefix().as_deref() == Some("app.kubernetes.io");
```

Using `as_deref()` on an `Option<&KeyPrefix>` will return `&KeyPrefix` instead of `&str`. A simplified code snippet demonstrates this rather nicely ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=57010767763cf731d523ef0b9809231e)).

Manually implementing `Deref for &T` collides with a blanket implemention in `core`. Any idea what we can do in this case?

rust/crd/src/lib.rs
NickLarsenNZ: @Techassi rightly mentioned (in a DM) that the key/value fields would be better suited in the underlying error.

So I'll just name the variants appropriately for now.

NickLarsenNZ: https://github.com/stackabletech/operator-rs/pull/726

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: @nightkr I was having a look through this code a bit closer.
Currently we error out if there are no labels found on the pod.

You are suggesting to not throw and error, and simply return an empty set of labels. Sounds reasonable at first glance.

But these labels are then filtered (for `app.kubernetes.io/*`) and used as labels for a service.

- [ ] Should we continue to error out as was already done?
- [ ] Return an empty set of labels so the service ends up being created without any of the pods `app.kubernetes.io/*` labels?

I would prefer the first option for this PR.

NickLarsenNZ: Ah ok, sorry I thought they were used as a selector, but they are just used as labels on the service. So not too dangerous, but maybe we want to handle a weird state where the pod has no labels, and then also if there are no `app.kubernetes.io/*` labels?

rust/crd/src/lib.rs
nightkr: Agreed that op-rs is the better place for fleshing out why it's invalid, but I think hdfs-op is the only place that can know where those values came from or what we're trying to do, which is IMO as important to mention.

But we can do that later.

rust/operator-binary/src/pod_svc_controller.rs
nightkr: I wouldn't say that's the responsibility of the pod svc controller to handle.

rust/operator-binary/src/pod_svc_controller.rs
nightkr: `.map(Deref::deref)` should work (or `.map(|x| &**x)` if you prefer sigil soup :P).

rust/operator-binary/src/kerberos.rs
nightkr: There's also https://github.com/stackabletech/hdfs-operator/pull/454/files#diff-a4441abfc6a76cee374b7e2b80a8a76901cbff6d19877be5b13de19453d4604aR246.

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: Oh cool, yeah this works (double deref until the next operator-rs release):

```rust
            label
                .key()
                .prefix()
                .map(std::ops::Deref::deref)
                .map(std::ops::Deref::deref)
                == Some(APP_KUBERNETES_LABEL_BASE)
```

> if you prefer sigil soup

I've already eaten, thanks :P

rust/operator-binary/src/pod_svc_controller.rs
NickLarsenNZ: So, @Techassi how can I do `.as_ref_or_else(BTreeMap::default)` but for `Labels`?

I've tried `Labels::default`, `Labels::new`. The error is posted a few comments above this.

nightkr: Okay, let's not block on this. It's not a regression and will die soon in #450 anyway.

rust/crd/src/lib.rs
NickLarsenNZ: Resolved in af6efe38c77bd6f573448d0892f452f71e2588bf

rust/operator-binary/src/kerberos.rs
NickLarsenNZ: Ahh :facepalm:
Maybe I should just name it `is_not_supported`

NickLarsenNZ: Fixed (?) in 2353cc71c91ee3a18147b11ee313c63cdc12e2ae

rust/operator-binary/src/hdfs_controller.rs
```diff
@@ -69,105 +72,100 @@ const DOCKER_IMAGE_BASE_NAME: &str = "hadoop";
 #[derive(Snafu, Debug, EnumDiscriminants)]
 #[strum_discriminants(derive(IntoStaticStr))]
 pub enum Error {
```
Techassi, line 74: note: We might want to reduce the number of variants in this error by splitting functionality up into modules. But this is out of scope for this PR.

rust/operator-binary/src/kerberos.rs
```diff
@@ -147,11 +154,11 @@ impl CoreSiteConfigBuilder {
 /// ```
 ///
 /// After we have switched to using the following principals everything worked without problems
-fn principal_host_part(hdfs: &HdfsCluster) -> Result<String, Error> {
+fn principal_host_part(hdfs: &HdfsCluster) -> Result<String> {
     let hdfs_name = hdfs.name_any();
     let hdfs_namespace = hdfs
         .namespace_or_error()
-        .map_err(|_| Error::ObjectHasNoNamespace {
+        .with_context(|_| ObjectHasNoNamespaceSnafu {
```
Techassi, line 161: question: Why can't we use `.context()` here? No complaint, just me asking for a short explanation.

Techassi (approved):
Great work, really like how you cleaned up the errors and the different builders.

rust/operator-binary/src/kerberos.rs
NickLarsenNZ: _Please select multiple lines so the context is clearer_

https://github.com/stackabletech/hdfs-operator/blob/0e928ae0771bc43e73b9e4f8f3174085c18d4716/rust/operator-binary/src/kerberos.rs#L161-L163

`context(SomeSnafu)` is evaluated immediately, meaning `ObjectRef::from_obj(hdfs)` and whatever it calls is also run immediately.

`with_context(f -> SomeSnafu)` on the other hand is evaluated when `f` is called, which is only in the error case.

So it can prevent expensive computation.

The rue of thumb I have now been using is:
- simple Snafu with no extra fields, use `context()`
- otherwise, use `with_context()`

Techassi: Ah, good point - didn't cross my mind. Thank you for the explanation!

rust/operator-binary/src/hdfs_controller.rs
NickLarsenNZ: Yeah I agree, the larger error enums will tend to have groups of variants which should naturally lead to modules of errors.

NickLarsenNZ: We can also put the Result type alias into said modules, so it is still easy:

```rust
fn foo() -> mod_a::Result<Bloo> { ... }
fn bar() -> mod_b::Result<Blah> { ... }
```

NickLarsenNZ:
Merging, even though kuttl tests aren't working yet. We will fix forward.

PR #457, opened 2024-01-18
Remove `MergedConfig`

### LINKED ISSUE #456: Drop `MergedConfig` trait
* Move the common options into `struct CommonNodeConfig` instead of being duplicated across each type
* Add `#[serde(flatten)] common: CommonNodeConfig` into `{Journal,Name,Data}NodeConfig`
* Replace `MergedConfig` with `enum AnyNodeConfig { JournalNode(JournalNodeConfig), ... } impl AnyNodeConfig { fn common(&self) -> CommonNodeConfig { ... } }`

nightkr:
# Description

Fixes #456.

This is intended to solve two things:

1. Having to add every new config field to umpteen places, adding a lot of busywork to every change (and a lot of annoying indirection when trying to chase down the real definition of everything).
2. `MergedConfig` containing the union of all roles' options, making it unclear which fields available when (and the especially comical case where `resources` exists in all variants... but still has a separate accessor method for each).

Instead, this PR delegates actually-common options to the new `struct CommonNodeConfig` and introduces a new `enum AnyNodeConfig` which users can explicitly downcast from to access role-specific options. Common options are accessible immediately because `AnyNodeConfig` derefs to `CommonNodeConfig`.

This PR replaces all uses of `&dyn MergedConfig` with `&AnyNodeConfig` to minimize impact. In the future, it's intended that new code that is only relevant for a single role should use that role's config type directly.

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/deploy/helm/hdfs-operator/crds/crds.yaml
+++ b/deploy/helm/hdfs-operator/crds/crds.yaml
@@ -107,6 +107,7 @@ spec:
                       type: object
                     config:
                       default: {}
+                      description: Configuration options that are available for all roles.
                       properties:
                         affinity:
                           default:
@@ -3614,6 +3615,7 @@ spec:
                             type: object
                           config:
                             default: {}
+                            description: Configuration options that are available for all roles.
                             properties:
                               affinity:
                                 default:
@@ -7179,6 +7181,7 @@ spec:
                       type: object
                     config:
                       default: {}
+                      description: Configuration options that are available for all roles.
                       properties:
                         affinity:
                           default:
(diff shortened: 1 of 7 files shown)


razvan (dismissed):
nice. maybe an entry in the change log would be nice, even if it's not user facing.

razvan (approved):

nightkr:
I would argue that these externally invisible changes don't make sense to add, "refactored `MergedConfig`" isn't something that tells a user.. anything.

razvan:
This is probably a _discipline_ vs. _common sense_ discussion. I can get behind both arguments, just mentioned it because it's a checkbox for the reviewer.

PR #472, opened 2024-02-07
chore: Lower requested resources in the logging test

siegfriedweber:
# Description

The logging test creates the following pods:

* hdfs-test-runner-0
* hdfs-vector-aggregator-0
* test-hdfs-automatic-log-datanode-default-0
* test-hdfs-automatic-log-journalnode-default-0
* test-hdfs-automatic-log-namenode-default-0
* test-hdfs-automatic-log-namenode-default-1
* test-hdfs-custom-log-datanode-default-0
* test-hdfs-custom-log-journalnode-default-0
* test-hdfs-custom-log-namenode-default-0
* test-hdfs-custom-log-namenode-default-1
* test-zk-server-default-0

They request a lot of CPU resources, especially because a Vector container is started for every HDFS pod.

The logging test usually passes if the other test which is running in parallel, does not request too much CPU. But if two logging tests for different Hadoop versions are running in parallel then not all pods can be scheduled (on our AKS cluster) due to insufficient CPU resources. If pods from both tests cannot be scheduled then the tests are deadlocked and will fail both.

Failed test due to deadlock: https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/119/

The solution is to lower the requested resources.

This pull request lowers only the CPU requests for the HDFS containers because this can be easily done in the configuration. Lowering the requests for the Vector containers would require pod overrides (https://github.com/stackabletech/hdfs-operator/blob/23.11.0/rust/operator-binary/src/container.rs#L181-L187). However, this seems to be sufficient.

Successful test with lowered resources: https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/121/

### DIFF
--- a/tests/templates/kuttl/logging/03-install-hdfs-automatic-log.yaml.j2
+++ b/tests/templates/kuttl/logging/03-install-hdfs-automatic-log.yaml.j2
@@ -11,6 +11,13 @@ spec:
     vectorAggregatorConfigMapName: hdfs-vector-aggregator-discovery
     zookeeperConfigMapName: test-hdfs-automatic-log-znode
   nameNodes:
+    config:
+      resources:
+        memory:
+          limit: 256Mi
+        cpu:
+          max: 250m
+          min: 50m
     roleGroups:
       default:
         replicas: 2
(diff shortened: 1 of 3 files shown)


sbernauer (approved):
Thanks for the writeup!

PR #492, opened 2024-03-07
fix(getting started): use cluster-internal listener class

razvan:
Fixes this problem:

```
❯ kubectl exec -n default webhdfs-0 -- curl -s -XGET "http://simple-hdfs-namenode-default-0:9870/webhdfs/v1/testdata.txt?user.name=stackable&op=CREATE&noredirect=true"
command terminated with exit code 6
```


NickLarsenNZ (approved):
This works for me :rocket:

sbernauer:
Please always use something accessible from outside of k8s for getting started guides, created https://github.com/stackabletech/hdfs-operator/pull/493

PR #493, opened 2024-03-08
docs: Revert using cluster-internal for getting started guide

sbernauer:
# Description

Reverting parts of https://github.com/stackabletech/hdfs-operator/pull/492

```[tasklist]
# Author
- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs-style-guide).
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs-style-guide).
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


docs/modules/hdfs/examples/getting_started/hdfs.yaml
```diff
@@ -11,13 +11,13 @@ spec:
     dfsReplication: 1
   nameNodes:
     config:
-      listenerClass: cluster-internal
+      listenerClass: external-stable
```
razvan: why is this needed?

sbernauer: It is not strictly needed, but I would argue users want to able to access the stuff the just deployed ;)
And port-forwards IMHO should be a thing of last resort as they break all the time.

In this concrete HDFS case users accessing the WebUIs, for Trino and Superset it would be submitting queries or creating Dashboards,

razvan: I was confused when debugging  because this particular guide doesn't access HDFS from outside the cluster. Also the doc doesn't explain what the `listenerClass` even does (which is fine in this context) , that's why I changed it to the "safer" value.

sbernauer: I see. The default is actually `cluster-internal` for security reasons, so we need to set it explicitly in case we want external access.

Would you be in favor of internal or external access here?

razvan: > Would you be in favor of internal or external access here?

In the current state of the guide I would prefer "internal". If the guide starts explaining the listener class and/or adds instructions on how to access the cluster from outside, then yes, change it to "external".

sbernauer: I think we need some sort of strategy/guideline what we want to do, created https://github.com/stackabletech/decisions/issues/7. Happy about comments!

razvan (approved):
as discussed. thanks.

sbernauer:
Thanks!

PR #517, opened 2024-05-02
Allow dashes in port names (again)

### LINKED ISSUE #514: Port names with dashes are not supported in sidecar containers
## Port names with dashes are not supported in sidecar containers

The following example adds an OAuth proxy to the HDFS name nodes via `podOverrides`. A port is defined with a name that contains a dash.

```yaml
---
apiVersion: hdfs.stackable.tech/v1alpha1
kind: HdfsCluster
spec:
  nameNodes:
    podOverrides:
      spec:
        containers:
          - name: oauth-proxy
            image: quay.io/openshift/origin-oauth-proxy:4.12
            ports:
              - name: oauth-proxy
                containerPort: 8080
            ...
```

The namenode container uses this bash script to export the ports as environment variables:

```
if [[ -d /stackable/listener ]]; then
    export POD_ADDRESS=$(cat /stackable/listener/default-address/address)
    for i in /stackable/listener/default-address/ports/*; do
        export $(basename $i | tr a-z A-Z)_PORT="$(cat $i)"
    done
fi
```

This leads to the following error:

```
+ for i in /stackable/listener/default-address/ports/*
++ tr a-z A-Z
++ basename /stackable/listener/default-address/ports/oauth-proxy
++ cat /stackable/listener/default-address/ports/oauth-proxy
+ export OAUTH-PROXY_PORT=8080
/bin/bash: line 39: export: `OAUTH-PROXY_PORT=8080': not a valid identifier
```

The problem is that `-` is not allowed in an environment variable name.

A solution would be to convert the port name to a valid environment variable name. Attention must be paid because port names can start with a digit, e.g. `123-web`, whereas environment variables cannot.

Another solution would be to not export these ports.

nightkr:
# Description

Fixes #514

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs-style-guide).
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 - Revert changing the getting started script to use the listener class `cluster-internal` ([#492]) ([#493]).
+- Fix HDFS pods crashing on launch when any port names contain dashes ([#517]).
+
+[#493]: https://github.com/stackabletech/hdfs-operator/pull/493
+[#517]: https://github.com/stackabletech/hdfs-operator/pull/517
 
 ## [24.3.0] - 2024-03-20
 
@@ -49,7 +53,6 @@ All notable changes to this project will be documented in this file.
 [#475]: https://github.com/stackabletech/hdfs-operator/pull/475
 [#491]: https://github.com/stackabletech/hdfs-operator/pull/491
 [#492]: https://github.com/stackabletech/hdfs-operator/pull/492
-[#493]: https://github.com/stackabletech/hdfs-operator/pull/493
 [#495]: https://github.com/stackabletech/hdfs-operator/pull/495
 [#499]: https://github.com/stackabletech/hdfs-operator/pull/499
 
--- a/rust/operator-binary/src/container.rs
+++ b/rust/operator-binary/src/container.rs
@@ -525,7 +525,7 @@ prepare_signal_handlers
 if [[ -d {LISTENER_VOLUME_DIR} ]]; then
     export POD_ADDRESS=$(cat {LISTENER_VOLUME_DIR}/default-address/address)
     for i in {LISTENER_VOLUME_DIR}/default-address/ports/*; do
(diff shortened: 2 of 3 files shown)


sbernauer (approved):
Thanks for fixing! All other cases should be handled

> As with Kubernetes [names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names) in general, names for ports must only contain lowercase alphanumeric characters and -. Port names must also start and end with an alphanumeric character.

Just for the record from https://github.com/stackabletech/hdfs-operator/issues/514#issuecomment-2090601235

> I'd argue that 123-abc is far less common, and while switching to a PORT_ prefix would fix it, I'm not sure it's worth the compatibility break.

sbernauer:
Thinking about it, would anything else than `rust/operator-binary/src/hdfs_controller.rs` be affected by adding a `PORT_` prefix?
We set e.g. `dfs.datanode.registered.port` to `${env.DATA_PORT}`, shouldn't his be unnoticed by customers?

nightkr:
(Merging since digit-prefixed ports are a separate fix anyway, but can continue discussing in here or make a new issue)

>  We set e.g. dfs.datanode.registered.port to ${env.DATA_PORT}, shouldn't his be unnoticed by customers?

In the default configuration, no. But they can still use `${env.}` references in their own config overrides, which _would_ be broken by it.

nightkr:
It would definitely have been better to do it that way from the start. Now we have the tradeoff to make.

sbernauer:
I think the likelihood of someone doing that is very very unlikely and I personally would prefer to be safe in the long run

nightkr:
To be clear, the options here are:

- Break previously-working configurations during routine upgrades, with an unclear cause or path forward
- Break when applying a new configuration that trips over the edge case, which can easily be reverted

sbernauer:
Honestly I don't care to much, I leave it up to you

PR #553, opened 2024-07-17
docs(getting-started): update stackablectl op install output

NickLarsenNZ:
```[tasklist]
### Run getting_started.sh via
- [x] `stackablectl`
- [x] `helm`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<!-- markdownlint-disable MD041 -->
+<!-- markdownlint-disable MD041 --><!-- markdownlint-disable MD041 -->
 <p align="center">
   <img width="150" src="./.readme/static/borrowed/Icon_Stackable.svg" alt="Stackable Logo"/>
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-<!-- markdownlint-disable MD041 -->
+<!-- markdownlint-disable MD041 --><!-- markdownlint-disable MD041 -->
```
maltesander, line 1: Why twice :) ?

NickLarsenNZ: Yup. Because the there are two files. The main one and the template partial.
We could alternatively just disable the rule completely (in operator-templating).

maltesander: Ok ill approve if its expected then :)

maltesander (dismissed):
LGTM!

maltesander (approved):
LGTM!

PR #562, opened 2024-07-24
fix: Adjust getting-started script for release 24.7.0

Techassi:
Part of https://github.com/stackabletech/issues/issues/613

- [x] Cherry-pick commit into release branch.

PR #563, opened 2024-07-26
fix: Adjust getting-started script for release 24.7.0

Techassi:
Part of #562


sbernauer (approved):

Techassi:
Also see https://github.com/stackabletech/druid-operator/pull/602#issuecomment-2250244293

sbernauer:
Jep, I saw the failing check and think we should ignore it

PR #571, opened 2024-08-02
Rolling HDFS upgrade

### LINKED ISSUE #362: Support [rolling] upgrade of HDFS
As of 23.4, when you upgrade your hdfs, e.g. 3.2.2 -> 3.3.4 you run into the error
```
2023-06-15 06:21:23,060 ERROR namenode.NameNode (NameNode.java:main(1839)) - Failed to start namenode.                                                                                 
java.io.IOException:                                                                                                                                                                   
File system image contains an old layout version -65.                                                                                                                                  
An upgrade to version -66 is required.                                                                                                                                                 
Please restart NameNode with the "-rollingUpgrade started" option if a rolling upgrade is already started; or restart NameNode with the "-upgrade" option to start a new upgrade.
```
Ideally we should start a rolling upgrade of all components. Currently you simply cannot upgrade your HDFS without hacking stuff (e.g. no cliOverrides to add `-upgrade` or similar)

Edit from the past: At least the upgrade 3.3.4 -> 3.3.6 worked

nightkr:
# Description

Fixes #362, currently implements the "manual" process from https://github.com/stackabletech/hdfs-operator/issues/362#issuecomment-2258312275.

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
...
+  Block Pool ID: BP-841432641-10.244.0.29-1722612757853
+     Start Time: Fri Aug 02 15:49:12 GMT 2024 (=1722613752341)
+  Finalize Time: Fri Aug 02 15:58:39 GMT 2024 (=1722614319854)
+----
+
+// We can't safely automate this, because finalize is asynchronous and doesn't tell us whether all NameNodes have even received the request to finalize.
+
+WARNING: Please ensure that all NameNodes are running and available before proceeding. NameNodes that have not finalized yet will crash on launch when taken out of upgrade mode.
+
+Finally, the operator and cluster should be taken out of upgrade mode, by marking the HdfsCluster as upgraded to the new version:
```
nightkr: Unclear terminology here.. we essentially have three different "upgrade modes" that we need to keep track of:

1. HDFS cluster upgrade mode (managed by `dfsadmin`)
2. NameNode upgrade mode (allows the NameNode to launch despite not matching the version of its local fsimage)
3. HdfsCluster stacklet upgrade mode (*currently* manages 2 but not 1, ideally eventually it would manage both)

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
+= Upgrading HDFS
+
+IMPORTANT: HDFS upgrades are experimental, and details may change at any time
+
+HDFS currently requires a manual process to upgrade. This guide will take you through an example case, upgrading an example cluster (from our xref:getting_started/index.adoc[Getting Started] guide) from HDFS 3.3.6 to 3.4.0.
+
+== Preparing HDFS
+
+HDFS must be configured to initiate the upgrade process. To do this, run the following commands in a HDFS superuser environment
+(either a client configured with a superuser account, or from inside NameNode pod):
```
NickLarsenNZ:
```suggestion
HDFS must be configured to initiate the upgrade process. To do this, run the following commands in an HDFS superuser environment
(either a client configured with a superuser account, or from inside a NameNode pod):
```

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+// This could be automated by the operator, but dfsadmin does not have good machine-readable output.
+// It *can* be queried over JMX, but we're not so lucky for finalization.
```
NickLarsenNZ, line 30: Dare I ask if we should write some Java to do whatever it is that dfsadmin would have been doing :woozy_face:?

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
...
+     Start Time: Fri Aug 02 15:49:12 GMT 2024 (=1722613752341)
+  Finalize Time: <NOT FINALIZED>
+
+$ # It should look like this once ready
+$ hdfs dfsadmin -rollingUpgrade query
+QUERY rolling upgrade ...
+Proceed with rolling upgrade:
+  Block Pool ID: BP-841432641-10.244.0.29-1722612757853
+     Start Time: Fri Aug 02 15:49:12 GMT 2024 (=1722613752341)
+  Finalize Time: <NOT FINALIZED>
```
NickLarsenNZ: I think the difference when ready should be made clearer. From what can tell, one line of text changes from "Preparing for upgrade. Data is being saved for rollback." to "Proceed with rolling upgrade:".

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
...
+----
+$ hdfs dfsadmin -rollingUpgrade finalize
+FINALIZE rolling upgrade ...
+Rolling upgrade is finalized.
+  Block Pool ID: BP-841432641-10.244.0.29-1722612757853
+     Start Time: Fri Aug 02 15:49:12 GMT 2024 (=1722613752341)
+  Finalize Time: Fri Aug 02 15:58:39 GMT 2024 (=1722614319854)
+----
+
+// We can't safely automate this, because finalize is asynchronous and doesn't tell us whether all NameNodes have even received the request to finalize.
```
NickLarsenNZ, line 93: Does that mean the finalized date appears even if there are still NameNodes that aren't finalized yet?

rust/operator-binary/src/container.rs
```diff
@@ -566,11 +566,16 @@ if [[ -d {LISTENER_VOLUME_DIR} ]]; then
...
+{hadoop_home}/bin/hdfs {role} {upgrade_args} &
 wait_for_termination $!
 {create_vector_shutdown_file_command}
 "#,
                     hadoop_home = Self::HADOOP_HOME,
+                    upgrade_args = if hdfs.is_upgrading() && *role == HdfsRole::NameNode {
+                        "-rollingUpgrade started"
+                    } else {
+                        ""
+                    },
```
NickLarsenNZ: **nit (non-blocker)**: This arg is getting a bit complex. It might be better to move it above as `let upgrade_args = ...`, and let format capture it.

rust/crd/src/lib.rs
```diff
@@ -802,6 +804,15 @@ impl HdfsCluster {
         Ok(result)
     }
 
+    pub fn is_upgrading(&self) -> bool {
+        self.status
+            .as_ref()
+            .and_then(|status| status.deployed_product_version.as_deref())
+            .map_or(false, |deployed_version| {
+                deployed_version != self.spec.image.product_version()
+            })
```
NickLarsenNZ: I like this style :)

rust/operator-binary/src/hdfs_controller.rs
```diff
@@ -323,10 +323,15 @@ pub async fn reconcile_hdfs(hdfs: Arc<HdfsCluster>, ctx: Arc<Ctx>) -> HdfsOperat
...
     let mut ss_cond_builder = StatefulSetConditionBuilder::default();
 
-    for (role_name, group_config) in validated_config.iter() {
-        let role: HdfsRole = HdfsRole::from_str(role_name).with_context(|_| InvalidRoleSnafu {
-            role: role_name.to_string(),
-        })?;
+    let mut deploy_done = true;
+
+    // Roles must be deployed in order during rolling upgrades
+    'roles: for role in HdfsRole::iter() {
```
NickLarsenNZ: I'm unsure what `'roles:` means here. Is it some magic syntax? The checks seemed to pass

rust/operator-binary/src/hdfs_controller.rs
```diff
@@ -408,14 +413,26 @@ pub async fn reconcile_hdfs(hdfs: Arc<HdfsCluster>, ctx: Arc<Ctx>) -> HdfsOperat
...
+                })?;
+            ss_cond_builder.add(deployed_rg_statefulset.clone());
+            if hdfs.is_upgrading() {
+                // When upgrading, ensure that each role is upgraded before moving on to the next as recommended by
+                // https://hadoop.apache.org/docs/r3.4.0/hadoop-project-dist/hadoop-hdfs/HdfsRollingUpgrade.html#Upgrading_Non-Federated_Clusters
+                if let Err(reason) = check_all_replicas_updated(&deployed_rg_statefulset) {
+                    tracing::info!(
+                        object = %ObjectRef::from_obj(&deployed_rg_statefulset),
+                        reason = &reason as &dyn std::error::Error,
+                        "rolegroup is still upgrading, waiting..."
```
NickLarsenNZ: **thought(non-blocker)**: I think we should use OpenTelemetry semantic conventions here, for example [`k8s.statefulset.name`](https://opentelemetry.io/docs/specs/semconv/resource/k8s/#statefulset) and [`k8s.namespace.name`](https://opentelemetry.io/docs/specs/semconv/resource/k8s/#namespace). But I think `ObjectRef::from_obj` gives both of those in one which makes this idea a bit more difficult.

rust/operator-binary/src/hdfs_controller.rs
```diff
@@ -459,12 +476,31 @@ pub async fn reconcile_hdfs(hdfs: Arc<HdfsCluster>, ctx: Arc<Ctx>) -> HdfsOperat
...
+        //         .and_then(|status| status.deployed_product_version.clone())
+        // },
     };
 
-    cluster_resources
-        .delete_orphaned_resources(client)
-        .await
-        .context(DeleteOrphanedResourcesSnafu)?;
+    // During upgrades we do partial deployments, we don't want to garbage collect after those
+    // since we *will* redeploy (or properly orphan) the remaining resources layer.
```
NickLarsenNZ:
```suggestion
    // During upgrades we do partial deployments, we don't want to garbage collect after those
    // since we *will* redeploy (or properly orphan) the remaining resources later.
```

NickLarsenNZ (dismissed):
LGTM, just a few typos and non-blocking thoughts and questions

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
nightkr: I was thinking about it, but [`HdfsAdmin`](https://hadoop.apache.org/docs/r3.4.0/api/org/apache/hadoop/hdfs/client/HdfsAdmin.html) doesn't expose the API to manage it. From what I remember, DFSClient *does* but isn't part of the public client API. I'm not saying we shouldn't ever do it, but I think it makes sense to break it out as future work.

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
nightkr: Yes! As far as I can tell, there is no way to query for finalization status, beyond "well, if all pods have been ready for a few seconds after finalization then you're prooobably fine". Isn't HDFS fun?

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
nightkr: Agreed, happy to take suggestions for how to display this better.

rust/operator-binary/src/hdfs_controller.rs
nightkr: Fair point. `object` is also probably not great here either, since it overlaps with the `object.ref` injected by the controller's span.

rust/operator-binary/src/hdfs_controller.rs
nightkr: It's a break label, so that we can `break 'roles;` out of the outer loop later on.

NickLarsenNZ: TIL! :+1: 

I know it from other languages, just never saw it in Rust.

rust/operator-binary/src/hdfs_controller.rs
NickLarsenNZ: We can update it later if we get some good ideas :D

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
NickLarsenNZ: If you mean the worst kind of fun, sure :D

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
NickLarsenNZ: I have made a suggestion to the hint. Feel free to edit it in the comment.

```suggestion
$ # It is safe to proceed when the output indicates so, like this:
$ hdfs dfsadmin -rollingUpgrade query
QUERY rolling upgrade ...
Proceed with rolling upgrade:
  Block Pool ID: BP-841432641-10.244.0.29-1722612757853
     Start Time: Fri Aug 02 15:49:12 GMT 2024 (=1722613752341)
  Finalize Time: <NOT FINALIZED>
```

NickLarsenNZ (dismissed):
LGTM

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
+= Upgrading HDFS
+
+IMPORTANT: HDFS upgrades are experimental, and details may change at any time
+
+HDFS currently requires a manual process to upgrade. This guide will take you through an example case, upgrading an example cluster (from our xref:getting_started/index.adoc[Getting Started] guide) from HDFS 3.3.6 to 3.4.0.
```
NickLarsenNZ, line 5: Should we mention anything about not using HDFS while the upgrade happens (just because of the point in time recovery for rollback?)

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
...
+QUERY rolling upgrade ...
+Proceed with rolling upgrade:
+  Block Pool ID: BP-841432641-10.244.0.29-1722612757853
+     Start Time: Fri Aug 02 15:49:12 GMT 2024 (=1722613752341)
+  Finalize Time: <NOT FINALIZED>
+----
+
+== Starting the upgrade
+
+Once ready, the HdfsCluster can be updated with the new product version:
```
NickLarsenNZ: Make it clear who/what needs to be ready.

```suggestion
Once HDFS is ready to upgrade, the HdfsCluster can be updated with the new product version:
```

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+[source,shell]
+----
+$ kubectl patch hdfs/simple-hdfs --patch '{"spec": {"image": {"productVersion": "3.4.0"}}}' --type=merge
+hdfscluster.hdfs.stackable.tech/simple-hdfs patched
+----
+
+Then wait until all pods are ready and running the new HDFS version.
```
NickLarsenNZ:
```suggestion
Then wait until all pods have restarted, are in the Ready state, and running the new HDFS version.
```

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+NOTE: Services will be upgraded in order: JournalNodes, then NameNodes, then DataNodes.
+
+== Finalizing the upgrade
+
+Once all HDFS pods are running the new version, the HDFS upgrade can be finalized (from the HDFS superuser environment):
```
NickLarsenNZ:
```suggestion
Once all HDFS pods are running the new version, the HDFS upgrade can be finalized (from the HDFS superuser environment as described in the preparation step):
```

rust/operator-binary/src/utils/statefulset.rs
sbernauer, line 1: It would be awesome if we could move this into operator-rs

nightkr: That's fair, I can prepare a PR. Do we want to block on that, or leave migrating to that implementation as future work?

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
NickLarsenNZ: Which terminology is unclear?

I've read through the whole doc from start to finish (hence new comments :sweat_smile:). The instructions seem clear to me.

nightkr: Mostly the two different upgrade modes, and what controls which.

rust/operator-binary/src/utils/statefulset.rs
nightkr: https://github.com/stackabletech/operator-rs/pull/833

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
nightkr: Fair point, I'm honestly not sure how HDFS behaves here. Maybe we should test that...

rust/operator-binary/src/container.rs
nightkr: eb19010953f10fd0397c84516e1382154903212a

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
NickLarsenNZ: I guess they are more _phases_ than _modes_. All of them required, some parts manual, some parts done by the operator.

I think that is quite clear in the instructions, but I guess if it needs more clarity, the text at the top could be reworked?

`HDFS currently requires a manual process to upgrade.`

Could be

`HDFS currently requires manual steps to both begin and finalize process. The HDFS Operator will take care of bringing up the NameNodes in a filesystem compatibility mode` (or whatever the correct way of saying that is).

nightkr: > I guess they are more phases than modes.

Sort of. Phases from the POV of the overall process, but modes from the POV of each daemon.

> HDFS currently requires manual steps to both begin and finalize process. The HDFS Operator will take care of bringing up the NameNodes in a filesystem compatibility mode (or whatever the correct way of saying that is).

Okay, I quite like compatibility mode for the NameNodes.

rust/operator-binary/src/utils/statefulset.rs
sbernauer: If you are fine with it I would prefer doing it right the first time. But in case it's urgent feel free to merge anyway. At least I can promise a quick review :see_no_evil:

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
sbernauer: Makes sense! Speaking of rollback, is this something we want to test and support as well?
I'd be fine not to do for now, in that case can we please document that we don't support rollback? ^^

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
nightkr: f0df2b7f49e5c9a32f9c701c6d1c342ec2127f0e

rust/operator-binary/src/utils/statefulset.rs
nightkr: 49cf9d9061bd59691bd2a5957c1a78cc591bb399

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
```diff
@@ -0,0 +1,87 @@
...
+  Block Pool ID: BP-841432641-10.244.0.29-1722612757853
+     Start Time: Fri Aug 02 15:49:12 GMT 2024 (=1722613752341)
+  Finalize Time: Fri Aug 02 15:58:39 GMT 2024 (=1722614319854)
+----
+
+// We can't safely automate this, because finalize is asynchronous and doesn't tell us whether all NameNodes have even received the request to finalize.
+
+WARNING: Please ensure that all NameNodes are running and available before proceeding. NameNodes that have not finalized yet will crash on launch when taken out of compatibility mode.
+
+Finally, we mark the cluster as upgraded:
```
NickLarsenNZ: I think it should be worded imperitively, per [doc style guide](https://docs.stackable.tech/home/stable/contributor/docs/style-guide#_highlights)

```suggestion
Finally, mark the cluster as upgraded:
```

or

```suggestion
Finally, you can mark the cluster as upgraded:
```

nightkr:
Old CRD review vote comment, please ignore

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
nightkr: So, it seems like they differentiate between rollbacks (which roll back file data, require downtime) and downgrades (which preserves file data, possible in HA).

This PR supports downgrades, but does _not_ implement the ordering requirements to preserve HA. That would be possible to implement, but requires a few (additive) CRD changes. Rollbacks are a more complicated dance that I would break out into a separate issue.

nightkr: Added some documentation on downgrades to b24c25f43b5bfa82f4233531e62c34fa901c1c35, experimenting a bit with how complicated "proper" HA downgrades would be.

NickLarsenNZ: I'll leave this open then, but feel free to mark as resolved if the experiment doesn't yield the results you'd like.

NickLarsenNZ (dismissed):
LGTM

docs/modules/hdfs/pages/usage-guide/upgrading.adoc
nightkr: 10e5220f0ad70198196b4e429fe7400cd3cf231b

nightkr:
Moving back into RfC phase since 10e5220f0ad70198196b4e429fe7400cd3cf231b adds a new `.status.upgradingProductVersion` CRD field.

NickLarsenNZ:
> Moving back into RfC phase since 10e5220f0ad70198196b4e429fe7400cd3cf231b adds a new `.status.upgradingProductVersion` CRD field.

I didn't see the new field in the CRD YAML.

nightkr:
My bad, forgot to make crds. a9809badfd77953216894288e82701c0230a0791

NickLarsenNZ:
We could probably move it back to voting to speed it up.
I'll withhold my PR Approval until it's back out of RfC.

sbernauer:
TLDR

```rust
pub struct HdfsClusterStatus {
    /// The product version that the HDFS cluster is currently running.
    ///
    /// During upgrades, this field contains the *old* version.
    pub deployed_product_version: Option<String>,

    /// The product version that is currently being upgraded to, otherwise null.
    pub upgrading_product_version: Option<String>,
}
```

Which works for me. WDYT of using `upgrade_target_product_version` to make it a bit more clear? Or would you consider this too chatty?

nightkr:
Sounds like an improvement to me.

nightkr:
0604aa6a6d4c1474c457ab335bb3e64c705f16ec

nightkr:
Moving back into voting. Please vote on this comment.

sbernauer:
Voting finished, accepted

nightkr:
Current merge blockers are:

- https://github.com/stackabletech/hdfs-operator/pull/571#discussion_r1711645054 requires approval from @NickLarsenNZ 
- We need a new operator-rs release with https://github.com/stackabletech/operator-rs/pull/838

Cargo.toml
```diff
@@ -30,5 +30,4 @@ tracing-futures = { version = "0.2", features = ["futures-03"] }
 
 [patch."https://github.com/stackabletech/operator-rs.git"]
 #stackable-operator = { path = "../operator-rs/crates/stackable-operator" }
-#stackable-operator = { git = "https://github.com/stackabletech//operator-rs.git", branch = "main" }
-stackable-operator = { git = "https://github.com/stackabletech//operator-rs.git", branch = "revert/lazylock" }
+stackable-operator = { git = "https://github.com/stackabletech//operator-rs.git", branch = "main" }
```
NickLarsenNZ: Is there a specific commit we can pin it to? Or can we cut a release for what this needed from main?

rust/crd/src/lib.rs
```diff
@@ -804,13 +804,22 @@ impl HdfsCluster {
...
-            .map_or(false, |deployed_version| {
-                deployed_version != self.spec.image.product_version()
-            })
+    pub fn upgrade_state(&self) -> Option<UpgradeState> {
+        let status = self.status.as_ref()?;
+        let requested_version = self.spec.image.product_version();
+
+        if requested_version != status.deployed_product_version.as_deref()? {
+            // If we're requesting a different version than what is deployed, assume that we're upgrading.
+            // Could also be a downgrade to an older version, but we don't support downgrades after upgrade finalization.
```
NickLarsenNZ, line 821: Maybe I'm misunderstanding, but the new docs suggest we do support downgrading, just not rolling back.

rust/crd/src/lib.rs
```diff
@@ -804,13 +804,22 @@ impl HdfsCluster {
...
-            .and_then(|status| status.deployed_product_version.as_deref())
-            .map_or(false, |deployed_version| {
-                deployed_version != self.spec.image.product_version()
-            })
+    pub fn upgrade_state(&self) -> Option<UpgradeState> {
+        let status = self.status.as_ref()?;
+        let requested_version = self.spec.image.product_version();
+
+        if requested_version != status.deployed_product_version.as_deref()? {
+            // If we're requesting a different version than what is deployed, assume that we're upgrading.
```
NickLarsenNZ, line 820: **idea**: If HDFS respects semver, we could use the `semver` to determine if it is an upgrade or downgrade.

rust/crd/src/lib.rs
```diff
@@ -966,6 +975,14 @@ impl HdfsPodRef {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum UpgradeState {
+    /// The cluster is currently being upgraded to a new version.
+    Upgrading,
+    /// The cluster is currently being downgraded to the previous version.
+    Downgrading,
```
NickLarsenNZ: nit:

```suggestion
    /// The cluster is currently being upgraded to a new version.
    Upgrading,

    /// The cluster is currently being downgraded to the previous version.
    Downgrading,
```

NickLarsenNZ (commented):
Left some comments

rust/crd/src/lib.rs
nightkr: Basically, we support downgrading back *during* the upgrade, but only until you `rollingUpgrade -finalize`. After that you're "stuck". We also don't support something like "start at 3.3.4, upgrade to 3.3.6, abort and downgrade to 3.3.5", you would have to downgrade to 3.3.4 and *then* do a separate upgrade to 3.3.5.

rust/crd/src/lib.rs
nightkr: https://hadoop.apache.org/versioning.html looks like the same format as semver, but with slightly different bump criteria.

Either way, the question IMO is closer to "does this change actually _help_ anyone, and is that worth the complexity increase?", and the only improvement I really see in this case would be adding a guardrail to explicitly forbid the operation. But even that could be done simpler by a rule such as "if upgrade_target_product_version is not null then _only_ allow either it or deployed_product_version".

Cargo.toml
nightkr: Cargo automatically pins it in `Cargo.lock`. That said, we should wait for the release before merging this.

rust/crd/src/lib.rs
NickLarsenNZ: So basically in the current code, any change is considered an upgrade (even if the requested version is lower)?

If so, maybe the comment should be less precise?

```suggestion
            // If we're requesting a different version than what is deployed, assume that we're upgrading/downgrading.
```

---

**Edit**: I see the line below mentions downgrade.

rust/crd/src/lib.rs
NickLarsenNZ: > We also don't support something like "start at 3.3.4, upgrade to 3.3.6, abort and downgrade to 3.3.5", you would have to downgrade to 3.3.4 and then do a separate upgrade to 3.3.5.

Is there a guardrail for that case? Is it detected in the current code, or we just document what we do/don't support.

_I thought it would be quicker to ask than to rerun through the PR and guess_.

Cargo.toml
NickLarsenNZ: Yes true, but over time it could change when people do `cargo update`. So I meant pinned (as in locked) to a specific commit until it is release.

If you're ok with it, feel free to resolve this conversation.

NickLarsenNZ (dismissed):
Approving now so that when the last comment is answered/resolved, this can be merged.

Cargo.toml
NickLarsenNZ: https://github.com/stackabletech/operator-rs/releases/tag/stackable-operator-0.74.0 is released now

rust/crd/src/lib.rs
nightkr: There isn't currently, but I think it'd make sense to add one.

Cargo.toml
nightkr: 46eedee44f28e1b8ce2a007bdc92294d1693ce6b

rust/crd/src/lib.rs
nightkr: 2a25ff4071412a48843b0fa108f0029677863511

NickLarsenNZ (approved):
LGTM

PR #613, opened 2024-11-20
tests: Add OPA as a test dimension (release-24.11)

adwk67:
Replace hard-coded OPA version with a templated dimension.

There is a companion change for the main branch: #614.

PR #614, opened 2024-11-20
tests: Add OPA as a test dimension (main)

adwk67:
Replace hard-coded OPA version with a templated dimension.

There is a companion change for the release-24.11 branch: #613.


NickLarsenNZ (approved):
Nice find, thanks @adwk67!

PR #639, opened 2025-01-15
Remove useless default values for CRD

### LINKED ISSUE #626: Default Values in Helm Templates Cause ArgoCD Sync Issue
#### Problem  
Helm charts often include explicit default values for certain parameters in the rendered templates. However, when Kubernetes stores these resources in etcd, it removes these default values from the actual state. This discrepancy causes ArgoCD sync issues, as it detects differences between the desired state (explicit defaults) and the actual state (defaults removed).

For example, in the provided ArgoCD configuration file, `jqPathExpressions` are used to ignore these discrepancies:  

```yaml
ignoreDifferences:
- group: "apiextensions.k8s.io"
  kind: "CustomResourceDefinition"
  jqPathExpressions:
  - .spec.names.categories | select(. == [])
  - .spec.names.shortNames | select(. == [])
  - .spec.versions[].additionalPrinterColumns | select(. == [])
```

While this workaround resolves sync errors, it adds complexity to the configuration.

In addition, here is the full ArgoCD file which allow to install a bunck of stackable operators:
```yaml
{{ if and .Values.e2e.enabled .Values.stackable.enabled }}
{{- range $.Values.stackable.operators }}
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: "{{ . }}-operator"
  namespace: argocd
  labels:
    app.kubernetes.io/name: "{{ . }}-operator"
    app.kubernetes.io/part-of: fink
    app.kubernetes.io/component: operator
  finalizers:
  - resources-finalizer.argocd.argoproj.io
spec:
  project: default
  ignoreDifferences:
  - group: "apiextensions.k8s.io"
    kind: "CustomResourceDefinition"
    # Default values for these fields appears explicitely in desired state but are removed from actual state
    jqPathExpressions:
    - .spec.names.categories | select(. == [])
    - .spec.names.shortNames | select(. == [])
    - .spec.versions[].additionalPrinterColumns | select(. == [])
  source:
    chart: "{{ . }}-operator"
    repoURL: https://repo.stackable.tech/repository/helm-stable/
    targetRevision: 24.11.0
    helm:
      releaseName: "{{ . }}-operator"
      valuesObject:
        resources:
          requests:
            cpu: "0"
            memory: "0"
  destination:
    server: "https://kubernetes.default.svc"
    namespace: stackable-operators
  syncPolicy:
    syncOptions:
      - CreateNamespace=true
      - ServerSideApply=true
      - RespectIgnoreDifferences=true
---
{{- end }}
{{ end }}
```

#### Desired Behavior  
Helm templates should avoid explicitly rendering default values that Kubernetes automatically assigns. This would eliminate the need for `jqPathExpressions` in ArgoCD configurations, simplifying setup and maintenance.

#### Proposed Solution  
- Update the Helm chart templates to conditionally omit parameters that match Kubernetes' default values.  
  For instance:  
  ```yaml
  {{- if .Values.someParameter }}
  someField: {{ .Values.someParameter }}
  {{- end }}
  ```
- Ensure that the templates only render non-default values when explicitly provided.

#### Benefits  
- **Simplified ArgoCD Configurations**: Removing the need for `jqPathExpressions` in the `ignoreDifferences` section.  
- **Improved Sync Behavior**: Avoid unnecessary diffs between desired and actual states.  
- **Better Alignment with Kubernetes Practices**: Align Helm templates with Kubernetes behavior regarding default values.  

#### Impact  
This change would enhance compatibility with ArgoCD and improve usability for teams relying on GitOps workflows to manage Helm-based deployments.  

#### Additional Context  
For example, with this change, the provided ArgoCD configuration could omit the `ignoreDifferences` block, as the sync issues caused by explicit defaults would no longer occur.

fjammes:
# Description

Removed default values from CRDs as they were causing issues with synchronization in ArgoCD. When Kubernetes stored the CRDs inside etcd, it stripped fields with default values, which led to ArgoCD being unable to reconcile the changes properly. By eliminating default values, we ensure that the synchronization process in ArgoCD works without issues.

- Not all of these items are applicable to all PRs; the author should update this template to only leave the boxes that are relevant.
- Please make sure all these things are done and tick the boxes.

### Author

- [ ] CRD changes approved  
- [ ] CRD documentation for all fields, following the [[style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide)](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).  
- [x] Helm chart can be installed and deployed; operator works  
- [ ] Integration tests passed (for non-trivial changes)  

### Reviewer

- [ ] Documentation added or updated; follows the [[style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide)](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).  
- [ ] Changelog updated  

### Acceptance

- [ ] [[Roadmap](https://github.com/orgs/stackabletech/projects/25/views/1)](https://github.com/orgs/stackabletech/projects/25/views/1) has been updated  

--- 

Let me know if further refinements are needed!

### DIFF
--- a/deploy/helm/hdfs-operator/crds/crds.yaml
+++ b/deploy/helm/hdfs-operator/crds/crds.yaml
@@ -8,7 +8,6 @@ metadata:
 spec:
   group: hdfs.stackable.tech
   names:
-    categories: []
     kind: HdfsCluster
     plural: hdfsclusters
     shortNames:
@@ -16,8 +15,7 @@ spec:
     singular: hdfscluster
   scope: Namespaced
   versions:
-    - additionalPrinterColumns: []
-      name: v1alpha1
+    - name: v1alpha1
       schema:
         openAPIV3Schema:
           description: Auto-generated derived type for HdfsClusterSpec via `CustomResource`


nightkr:
As @maltesander mentioned in #626, those CRD files are generated. This is something that would need to be changed in [kube-derive](https://github.com/kube-rs/kube/blob/6a980c6ea50f2f3f4f2867d3df3fd77be00fca84/kube-derive/src/custom_resource.rs), which should then trickle down into all our operators.

*That said*, if Argo gets confused by explicitly initializing empty values then that sounds like a bug on their side.

fjammes:
Hi @nightkr , here is some detailed informations: this situation arises due to how Kubernetes handles default values in its API and the implications for tools like Argo CD. When a Kubernetes object is created, the API server automatically applies default values to fields that are not explicitly specified. However, these default values are not stored in etcd, the key-value store backing Kubernetes. Instead, **only explicitly set values are retained in etcd**.

On the other hand, Argo CD performs a comparison between the live state of the resource (as seen in etcd) and the desired state defined by Helm charts. Since the _stackable_ Helm templates for CRDs explicitly include a few fields with their default values, discrepancies arise during the comparison:

Helm-generated manifests: These contain explicitly defined default values.
Kubernetes live state: Does not include these default values, as they are not persisted in etcd.
This mismatch leads to Argo CD reporting a difference and blocking synchronization, even though the live state is functionally identical to the desired state due to Kubernetes' dynamic application of defaults.

This issue is unlikely to be resolved in Argo CD because it would require the tool to know all default values for every Kubernetes object, including CRDs. Handling this would mean Argo CD needs to fetch and interpret OpenAPI schemas for every resource and Kubernetes version, including custom CRDs. 

That's why I propose to remove the some default values from stackable CRDs, it is a simple and efficient short-term solution.

fjammes:
In addition, I have created a tiny kube-derive poc, do you think it is possible to configure it not to print fields with default values? https://github.com/k8s-school/kube-derive-poc

sbernauer:
kube `4.2.0` fixes this problem in https://github.com/kube-rs/kube/pull/2042.
SDP 26.11 will use that kube version (operator-rs version `0.115.0`, so the problem should be solved in the next release

PR #693, opened 2025-06-11
Add cipher suite (AES/CTR/NoPadding) per default

### LINKED ISSUE #680: Use AES by default for HDFS wire encryption
When Kerberos is enabled the operator automatically enables wire encryption by setting `dfs.data.transfer.protection` to "privacy" and `dfs.encrypt.data.transfer` to "true". By default the 3DES algorithm is used, but the faster and more secure AES algorithm is available. There is a marked performance improvement when AES is enabled compared to 3DES.

AES encryption can be enabled by adding the following property to `hdfs-site.xml`:
- dfs.encrypt.data.transfer.cipher.suite: "AES/CTR/NoPadding"

The key length can be defined using the property `dfs.encrypt.data.transfer.cipher.key.bitlength` to 128, 192 or 256, with 128 bit being the default. 

https://hadoop.apache.org/docs/r3.3.6/hadoop-project-dist/hadoop-common/SecureMode.html#Data_Encryption_on_Block_data_transfer.

maltesander:
## Description

fixes https://github.com/stackabletech/hdfs-operator/issues/680

The hdfs-operator now defaults to `AES/CTR/NoPadding` for `dfs.encrypt.data.transfer.cipher.suite` to improve security and performance

**hdfs-site.xml:**
```
<property>
  <name>dfs.data.transfer.protection</name>
  <value>privacy</value>
</property>
<property>
  <name>dfs.encrypt.data.transfer</name>
  <value>true</value>
</property>                                                                                  
<property>
  <name>dfs.encrypt.data.transfer.cipher.suite</name>
  <value>AES/CTR/NoPadding</value>                                                                                   
</property>
```

**Test:**
```
--- PASS: kuttl (608.24s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/kerberos_hadoop-3.4.1_zookeeper-latest-3.9.3_krb5-1.21.1_opa-1.4.2_kerberos-realm-PROD.MYCORP_kerberos-backend-mit_openshift-false (608.23s)
```

### DIFF
--- a/docs/modules/hdfs/pages/usage-guide/security.adoc
+++ b/docs/modules/hdfs/pages/usage-guide/security.adoc
@@ -33,7 +33,7 @@ The `kerberos.secretClass` is used to give HDFS the possibility to request keyta
 
 The `tlsSecretClass` is needed to request TLS certificates, used e.g. for the Web UIs.
 
-NOTE: The hdfs-operator defaults to `AES/CTR/NoPadding` for `dfs.encrypt.data.transfer.cipher.suite` with a default key length of 128 Bit. This can be changed using config overrides.
+NOTE: The hdfs-operator defaults to `AES/CTR/NoPadding` for `dfs.encrypt.data.transfer.cipher.suite`. This can be changed using config overrides.
 
 === 4. Verify that Kerberos authentication is required
 Use `stackablectl stacklet list` to get the endpoints where the HDFS namenodes are reachable.
--- a/docs/modules/hdfs/pages/usage-guide/security.adoc
+++ b/docs/modules/hdfs/pages/usage-guide/security.adoc
@@ -33,7 +33,7 @@ The `kerberos.secretClass` is used to give HDFS the possibility to request keyta
 
 The `tlsSecretClass` is needed to request TLS certificates, used e.g. for the Web UIs.
 
-NOTE: The hdfs-operator uses the cipher suite `AES/CTR/NoPadding` with a 128 Bit key per default. This can be changed using config overrides.
+NOTE: The hdfs-operator defaults to `AES/CTR/NoPadding` for `dfs.encrypt.data.transfer.cipher.suite` with a default key length of 128 Bit. This can be changed using config overrides.
 
 === 4. Verify that Kerberos authentication is required
 Use `stackablectl stacklet list` to get the endpoints where the HDFS namenodes are reachable.
--- a/docs/modules/hdfs/pages/usage-guide/security.adoc
+++ b/docs/modules/hdfs/pages/usage-guide/security.adoc
@@ -33,6 +33,7 @@ The `kerberos.secretClass` is used to give HDFS the possibility to request keyta
 
 The `tlsSecretClass` is needed to request TLS certificates, used e.g. for the Web UIs.
 
+NOTE: The hdfs-operator uses the cipher suite `AES/CTR/NoPadding` with a 128 Bit key per default. This can be changed using config overrides.
 
 === 4. Verify that Kerberos authentication is required
 Use `stackablectl stacklet list` to get the endpoints where the HDFS namenodes are reachable.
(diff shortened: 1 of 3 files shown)


NickLarsenNZ (dismissed):
LGTM

Does this need a release-note? If so, it will also need the `scheduled-for` and `release-note` label.

maltesander:
> LGTM
> 
> Does this need a release-note? If so, it will also need the `scheduled-for` and `release-note` label.

Not sure, would let @lfrancke decide.

rust/operator-binary/src/security/kerberos.rs
```diff
@@ -52,6 +52,11 @@ impl HdfsSiteConfigBuilder {
     fn add_wire_encryption_settings(&mut self) -> &mut Self {
         self.add("dfs.data.transfer.protection", "privacy");
         self.add("dfs.encrypt.data.transfer", "true");
+        self.add(
+            "dfs.encrypt.data.transfer.cipher.suite",
+            "AES/CTR/NoPadding",
+        );
+        self.add("dfs.encrypt.data.transfer.cipher.key.bitlength", "128");
```
lfrancke: I don't think that's needed as it's the default anyway and I think I'd prefer keeping the default values so it moves with the times should Hadoop update its default.

@Jimvin - any opinions?

lfrancke:
To answer the question: I think it's worth a release note entry

rust/operator-binary/src/security/kerberos.rs
Jimvin: Agreed, 128 bit is the default and doesn't need to be set explicitly. If the default changes in the HDFS code then we ought to inherit that.

maltesander: done https://github.com/stackabletech/hdfs-operator/pull/693/commits/0fa4937045b62621426b946263ece965fc3f0789

maltesander:
Release note: 
```
The HDFS operator now defaults `dfs.encrypt.data.transfer.cipher.suite` to `AES/CTR/NoPadding` to improve security and performance.
```

docs/modules/hdfs/pages/usage-guide/security.adoc
```diff
@@ -33,6 +33,7 @@ The `kerberos.secretClass` is used to give HDFS the possibility to request keyta
 
 The `tlsSecretClass` is needed to request TLS certificates, used e.g. for the Web UIs.
 
+NOTE: The hdfs-operator defaults to `AES/CTR/NoPadding` for `dfs.encrypt.data.transfer.cipher.suite` with a default key length of 128 Bit. This can be changed using config overrides.
```
lfrancke: Maybe adjust this comment as well and remove the mention of the key length otherwise we're back to the state where we need to check again for every release

NickLarsenNZ: Now that we don't specify the default, we'll need to somehow ensure this statement is correct in future updates.

NickLarsenNZ (dismissed):
LGTM

docs/modules/hdfs/pages/usage-guide/security.adoc
maltesander: Hmm true, i would mention just uses the hadoop default or something then?

maltesander: Yeah decided to not mention that at all https://github.com/stackabletech/hdfs-operator/pull/693/commits/92d6c967e59f003e8f107a5139f2632f9492d28c

maltesander: Yeah decided to not mention that at all https://github.com/stackabletech/hdfs-operator/pull/693/commits/92d6c967e59f003e8f107a5139f2632f9492d28c

lfrancke (approved):

PR #696, opened 2025-06-19
feat: improve default configuration for reliability and performance

### LINKED ISSUE #685: Improve default configuration (reliability & performance)
## Description

As a user of HDFS I want settings (see below for details) enabled to improve reliability in the face of crashes even if it comes at the expense of some minor performance issues.

### dfs.datanode.sync.behind.writes & dfs.datanode.synconclose

Our main use-case for having HDFS in SDP is for HBase and we want it to be as reliable as possible. Currently there is a minor chance that data written to HDFS is not actually synced to disk even if a block has been closed. Users in HBase can control this explicitly for the WAL but for flushes and compactions I believe they can't as easily (if at all). In _theory_ HBase should be able to recover from these failures but that comes at a cost and there's always a risk. Enabling this behaviour causes HDFS to sync to disk as soon as possible.

### io.file.buffer.size

The default (4096) hasn't changed since 2009.

> The size of buffer for use in sequence files. The size of this buffer should probably be a multiple of hardware page size (4096 on Intel x86), and it determines how much data is buffered during read and write operations. Must be greater than zero.

I suggest changing this from the current default of 4k to 128k to allow for faster transfers.

### dfs.namenode.handler.count

Defaults to 10 since _at least_ 2011.

This controls the concurrent number of client connections (this includes DataNodes) to the NameNode. Ideally we'd scale this with the number of DataNodes but this would lead to restarts of the NameNode so I'm in favor of increasing the default from 10 to 50 and not make it dynamic. This should lead to better performance due to more concurrency.

There's also a setting `dfs.namenode.service.handler.count` which is purely for connections from DataNodes but that is only used if `dfs.namenode.servicerpc-address` is also set which we don't do

### dfs.datanode.handler.count

Defaults to 10 since _at least_ 2012.

This controls the concurrent number of client connections to the DataNodes. We have no idea how many clients there may be so it's hard to automate this but the default can definitely be raised I'd say. I suggest increaing the default from 10 to 50 as well. This should lead to better performance due to more concurrency especially with use-cases like HBase.

### dfs.namenode.replication.max-streams & dfs.namenode.replication.max-streams-hard-limit

Default to 2 and 4 respectively since around 2013. This controls the number of maximum replication "jobs" a NameNode assigns to a DataNode in a single heartbeat. Increasing this number will increase network usage during replication events but can lead to faster recovery.

We could make the defaults much higher (10/50 or so) but I'd prefer to do a minor increase instead not to overload any networks.
I suggest doubling these numbers: 4 & 8

There are others who set this much higher though: https://support.huawei.com/enterprise/en/knowledge/EKB1100015760

### dfs.datanode.max.transfer.threads

Defaults to 4096 and hasn't changed since at least 2011.

The number of threads used for actual data transfer, so not very CPU heavy but IO bound. This is why the number is relatively high. But todays Java and IO should be able to handle more so I suggest bumping this to 8192 for more performance/concurrency.

## Tasks

- [x] Set `io.file.buffer.size` to 131072 by default in core-site.xml
- [x] Set `dfs.datanode.sync.behind.writes` to `true` by default in hdfs-site.xml
- [x] Set `dfs.datanode.synconclose` to `true` by default in hdfs-site.xml
- [x] Set `dfs.namenode.handler.count` to 50 by default in hdfs-site.xml
- [x] Set `dfs.datanode.handler.count` to 50 by default in hdfs-site.xml
- [x] Set `dfs.namenode.replication.max-streams` to 4 by default in hdfs-site.xml
- [x] Set `dfs.namenode.replication.max-streams-hard-limit` to 8 by default in hdfs-site.xml
- [x] Set `dfs.datanode.max.transfer.threads` to 8192 by default in hdfs-site.xml

## Release Notes

hdfs: Various setting defaults have been updated for better performance and reliability ([#685](https://github.com/stackabletech/hdfs-operator/issues/685))

## References

- https://hadoop-hbase.blogspot.com/2013/07/protected-hbase-against-data-center.html
- https://issues.apache.org/jira/browse/HDFS-2465
- https://issues.apache.org/jira/browse/AMBARI-13066
- https://issues.apache.org/jira/browse/HDFS-1539
- https://issues.apache.org/jira/browse/ACCUMULO-1946

razvan:
## Description

Fixes #685

Ran the smoke test and confirmed that the props are in `hdfs-site.xml`.

CI https://testing.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/25/

### Author

- [x] CRD changes approved
- [x] Release note snippet added

### Reviewer

- [ ] Changelog updated

### Acceptance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to this project will be documented in this file.
   - Use `--console-log-format` (or `CONSOLE_LOG_FORMAT`) to set the format to `plain` (default) or `json`.
 - The operator now defaults to `AES/CTR/NoPadding` for `dfs.encrypt.data.transfer.cipher.suite` to improve security and performance ([#693]).
 - The built-in Prometheus servlet is now enabled and metrics are exposed under the `/prom` path of all UI services ([#695]).
-- Added several properties to `hdfs-site.xml` that improve general performance and reliability ([#696])
+- Added several properties to `hdfs-site.xml` and `core-site.xml` that improve general performance and reliability ([#696])
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
   - Use `--console-log-format` (or `CONSOLE_LOG_FORMAT`) to set the format to `plain` (default) or `json`.
 - The operator now defaults to `AES/CTR/NoPadding` for `dfs.encrypt.data.transfer.cipher.suite` to improve security and performance ([#693]).
 - The built-in Prometheus servlet is now enabled and metrics are exposed under the `/prom` path of all UI services ([#695]).
+- Added several properties to `hdfs-site.xml` that improve general performance and reliability ([#696])
 
 ### Changed
 
@@ -50,6 +51,7 @@ All notable changes to this project will be documented in this file.
 [#684]: https://github.com/stackabletech/hdfs-operator/pull/684
 [#693]: https://github.com/stackabletech/hdfs-operator/pull/693
 [#695]: https://github.com/stackabletech/hdfs-operator/pull/695
+[#696]: https://github.com/stackabletech/hdfs-operator/pull/696
 
 ## [25.3.0] - 2025-03-21
 
--- a/rust/operator-binary/src/hdfs_controller.rs
+++ b/rust/operator-binary/src/hdfs_controller.rs
@@ -674,9 +674,6 @@ fn rolegroup_config_map(
                     // Enabling this behavior causes HDFS to sync to disk as soon as possible.
                     .add("dfs.datanode.sync.behind.writes", "true")
                     .add("dfs.datanode.synconclose", "true")
-                    // The default (4096) hasn't changed since 2009.
-                    // Increase to 128k to allow for faster transfers.
-                    .add("io.file.buffer.size", "131072")
                     // Defaults to 10 since at least 2011.
                     // This controls the concurrent number of client connections (this includes DataNodes)
                     // to the NameNode. Ideally, we'd scale this with the number of DataNodes but this would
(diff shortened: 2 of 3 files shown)


rust/operator-binary/src/hdfs_controller.rs
```diff
@@ -664,7 +664,15 @@ fn rolegroup_config_map(
                     )
                     .add("dfs.datanode.registered.hostname", "${env.POD_ADDRESS}")
                     .add("dfs.datanode.registered.port", "${env.DATA_PORT}")
-                    .add("dfs.datanode.registered.ipc.port", "${env.IPC_PORT}");
+                    .add("dfs.datanode.registered.ipc.port", "${env.IPC_PORT}")
+                    .add("dfs.datanode.sync.behind.writes", "true")
```
lfrancke, line 675: May I suggest adding comments to each of those. I think you can copy & paste from the relevant issue a lot.

Just to make it easier to see later why/how we ended up with these numbers and values.

lfrancke (changes requested):

rust/operator-binary/src/hdfs_controller.rs
razvan: [done](https://github.com/stackabletech/hdfs-operator/pull/696/commits/0baead3aa281d057ee698a819be0a8f563d382fa)

lfrancke: Thank you. I like that!

lfrancke:
I found a problem in my original issue. I believe at least one of the properties actually needs to go into `core-site.xml` instead. I'll check.

lfrancke (changes requested):
`io.file.buffer.size` needs to go into `core-site.xml` instead

Sorry for the mistake on my end :(

razvan:
> `io.file.buffer.size` needs to go into `core-site.xml` instead

Sorry I was too lazy to check. [Moved](https://github.com/stackabletech/hdfs-operator/pull/696/commits/05e52b728ed2feeb5a45ce7a85f70e45206fda11).

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
   - Use `--console-log-format` (or `CONSOLE_LOG_FORMAT`) to set the format to `plain` (default) or `json`.
 - The operator now defaults to `AES/CTR/NoPadding` for `dfs.encrypt.data.transfer.cipher.suite` to improve security and performance ([#693]).
 - The built-in Prometheus servlet is now enabled and metrics are exposed under the `/prom` path of all UI services ([#695]).
+- Added several properties to `hdfs-site.xml` that improve general performance and reliability ([#696])
```
lfrancke: Needs to be adapted now as well, sorry :(

lfrancke (changes requested):

CHANGELOG.md
razvan: [updated](https://github.com/stackabletech/hdfs-operator/pull/696/commits/2930110355ae2b826797df918297c8b7ec53fbc7)

lfrancke (approved):

PR #747, opened 2026-02-03
feat: Add graceful shutdown

Techassi:
Part of https://github.com/stackabletech/issues/issues/803.

With this PR, all concurrent tasks are gracefully shutdown when a SIGTERM is received. Previously only parts of the tasks honoured the signal which caused Pods to be stuck and eventually force killed after a timeout by Kubernetes.


maltesander (approved):
LGTM!

PR #753, opened 2026-02-13
chore: Prepare CRD versioning

Techassi:
Part of https://github.com/stackabletech/issues/issues/808.

This prepares the operator for future CRD versioning by adding a conversion webhook and maintaining the CRD(s) on its own.

### Release Note

- The operator now maintains its own CRD(s) instead of relying on Helm as a deployment mechanism. Additionally, the conversion webhook is now running alongside the controller for future CRD versioning.


maltesander (changes requested):
LGTM, changelog missing.

Techassi:
I already added it, but forgot to push it :/

maltesander (approved):
LGTM!

PR #761, opened 2026-03-03
chore: Add demo/stack labels if applicable

xeniape:
## Description

Part of https://github.com/stackabletech/stackable-cockpit/issues/187
This PR adds the `stackable.tech/stack` and `stackable.tech/demo` labels to the various resources produced through the cluster reconciliation if the cluster contains these labels.

Changes to the `rust/operator-binary/src/crd/storage.rs` and `rust/operator-binary/src/event.rs` are unrelated, they were suggestions by `cargo clippy`.

In `rust/operator-binary/src/hdfs_controller.rs` I reordered the cluster_resources creation/addition to put related code closer together for readability. The order of the cluster_resources itself didn't change.

This PR would be a template for the other operators.

### Author

- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).

### Reviewer

- [ ] Changelog updated

### Acceptance


sbernauer:
I have to admit I'm a bit confused at what exactly I'm looking at here. In my world the operators don't know and care about concepts such as stacks and demos.

I though https://github.com/stackabletech/stackable-cockpit/issues/187 was about adding stack=foo and demo=bar labels to the objects *deployed by stackablectl*, such as the HdfsCluster. When uninstalling a stack/demo, stackablectl can use these to gather the list of resources to delete.
This way stackablectl deletes the HdfsCluster and the ownerReferences take care of deleting STS, PDBs, CMs, Listeners etc.

Note: The operators add a vendor=stackable label to all resources, if that is missing that totally needs fixing, and is reflected as the step "Go through all the operators and add stackable.tech/vendor=Stackable label to resources that still miss it" in https://github.com/stackabletech/stackable-cockpit/issues/187

xeniape:
> I have to admit I'm a bit confused at what exactly I'm looking at here. In my world the operators don't know and care about concepts such as stacks and demos.
> 
> I though [stackabletech/stackable-cockpit#187](https://github.com/stackabletech/stackable-cockpit/issues/187) was about adding stack=foo and demo=bar labels to the objects _deployed by stackablectl_, such as the HdfsCluster. When uninstalling a stack/demo, stackablectl can use these to gather the list of resources to delete. This way stackablectl deletes the HdfsCluster and the ownerReferences take care of deleting STS, PDBs, CMs, Listeners etc.
> 
> Note: The operators add a vendor=stackable label to all resources, if that is missing that totally needs fixing, and is reflected as the step "Go through all the operators and add stackable.tech/vendor=Stackable label to resources that still miss it" in [stackabletech/stackable-cockpit#187](https://github.com/stackabletech/stackable-cockpit/issues/187)

Open to any other approach, this is just my initial idea :)

Unfortunately PVCs and therefore any data and listener related Resources don't seem to be tracked by the list of resources to delete, which is the intended behavior according to the code docs. Maybe there is a better approach to labeling those but honestly that part of the code was a bit unclear for me.

The main labeling in this PR is for those PVCs so deletion of them is possible, the other places I just added them "since I was already there in the code" so resources part of the demo can directly be identified when looking at them / filtering.

Edit: Actually, maybe I can utilize https://github.com/stackabletech/issues/issues/712 in the demos? Didn't look into it that much yet how it works, do you think I could solve this problem with it?

maltesander:
Hmm i feel we should rather configure the operator (or probably the cluster CRDs) to add additional labels that might be added for this case? Will be another operator-rs rollout though?

xeniape:
Just an update:
- The solution in this PR is not appropriate, thanks for the feedback! I agree
- Using the objectOverrides wasn't possible, due to the bug in merging volumeClaimTemplates, @sbernauer put up an upstream fix https://github.com/Arnavion/k8s-openapi/pull/176
- Nevertheless, the better solution would be to just delete the whole demo namespace. I ruled that out first, because we narrowed down the scope of https://github.com/stackabletech/stackable-cockpit/issues/187 to not consider non-default namespaces. But after discussing with @sbernauer decided to implement that since it simplifies the deletion process by a lot (similar to kuttl)

Will close this PR and instead open up future ones in demos to make them available in non-default namespaces.

